PR #202, opened 2026-06-07
feat(parko-kirra): L2 / CERT-006 — diverse-shadow GovernorComparator

justinlooney:
Before L2 the GovernorComparator ran IDENTICAL redundancy — two KirraGovernor instances built from the same config. That catches random/transient faults and state divergence but CANNOT catch systematic implementation faults: a bug in KirraGovernor::evaluate manifests identically in both copies, the comparator sees no divergence, and the wrong answer passes through.

L2 closes this gap by introducing a structurally diverse second governor (DiverseKirraGovernor) that enforces the SAME safety properties via deliberately different computation. A systematic implementation bug in one is unlikely to manifest identically in the other.

** Status: DRAFT — pending formal safety-engineer review. ** See docs/safety/COMPARATOR_DIVERSITY.md for the diversity argument, the fault class covered, and the explicit limit: spec-level systematic faults shared by both implementations are NOT covered (would require N-version reimplementation).

Decision: Option A — structural/algorithmic diversity. Tractable now, catches implementation-level systematic faults (the most common class). Option B (full N-version from spec only) is noted as the stronger-but-later step.

Concrete structural differences (the diversity argument; see the COMPARATOR_DIVERSITY.md doc §3 for the full table):

  1. Verdict composition — primary uses an early-return cascade; diverse builds per-axis AxisVerdict { Allow / Clamp / Deny } flags first, then composes into EnforcementAction at the end.
  2. Linear axis — primary calls validate_vehicle_command (kernel SDK); diverse reimplements P0-P4 inline with different algebra:
     - NaN guard: `is_nan() || is_infinite()` vs `!is_finite()`
     - Acceleration check: multiplicative `dv >= a*dt + ε` vs divisive `dv/dt > a + ε` - Clamp idiom: `v_max.copysign(v)` vs `v_max * v.signum()` - effective_max_speed: two `.min()` calls vs `match Option` - Skips steering (P5) / lateral-accel (P6) entirely (steering is hardcoded 0 in the bridge)
  3. Angular axis — primary calls AngularVelocityBound::omega_max; diverse recomputes inline with rearranged algebra: - Rollover: `(g/v)·(t/(2h))` vs `g·t/(2·h·v)` - Sweep: `v_edge * r_extent.recip()` vs `v_edge / r_extent` - v=0 floor: `if v < FLOOR` (mask) vs `if v >= FLOOR` (guard) - Fold: array + `.iter().fold(INF, min)` vs chained `.min().min()`
  4. MRC profile — diverse decomposes the two axes into independent helpers + a tuple match (vs primary's fused both-axes function).

Code:
  - New: parko/crates/parko-kirra/src/diverse.rs (~340 lines). DiverseKirraGovernor struct + impl SafetyGovernor. Methods diverse_linear_check, diverse_angular_check, diverse_omega_max, diverse_mrc, compose. 5 unit tests in diverse::tests covering the per-axis paths.
  - Modified: parko/crates/parko-kirra/src/comparator.rs:
    - GovernorComparator's primary/shadow fields changed from KirraGovernor to Box<dyn ComparatorGovernor + Send + Sync>. ComparatorGovernor = SafetyGovernor + RssAware (new local trait so the comparator can forward update_rss_state to a heterogeneous shadow).
    - Existing constructors (new / with_sink) keep their back-compat signature taking two KirraGovernor; new wraps internally.
    - New constructors: new_diverse(KirraGovernor, DiverseKirraGovernor)
      + with_diverse_shadow_and_sink(...).
    - update_rss_state dispatches via the RssAware trait method.

Tests (new in comparator::diversity_tests):

AGREEMENT (the critical correctness property):
  - diverse_agrees_with_primary_across_input_regimes 30 hand-built inputs spanning Allow / each Clamp* / Deny and each posture (Nominal / Degraded / LockedOut). Assertion: the comparator's own actions_diverge(p, d, .., COMPARATOR_TOLERANCE) == false for every input. A regression introducing false divergence on a valid input would fail this test.
  - diverse_agrees_with_primary_on_fail_closed_paths 6 fail-closed paths (NaN / Inf on each axis, zero / negative dt). Both governors must reach equivalent Deny.

DETECTION (demonstrable):
  - comparator_detects_fault_injected_shadow_disagreement Pair primary KirraGovernor with an AllowAlwaysFaulty shadow that always returns Allow. Send a 50 m/s command the primary clamps. Assertions: (a) reconciled effective linear < 50 (faulty Allow did NOT pass through), (b) InMemoryDivergenceSink recorded at least one DivergenceEvent.
  - sustained_divergence_escalates_to_locked_out Hammer divergent command 200 times; assert the leaky-bucket accumulator kicks in and a divergence event lands.

Pre-existing tests still pass: 8 comparator tests + 24 KirraGovernor tests + 12 angular_bound tests + posture tests all green.

Doc: docs/safety/COMPARATOR_DIVERSITY.md (~280 lines):
  - Prominent DRAFT — pending formal safety-engineer review banner.
  - The three orthogonal claims (correctness / diversity / detection) — each with its own justification kind.
  - §3 concrete structural-differences table by property.
  - §4 fault class COVERED (with rationale per row) AND the explicit LIMIT (spec-level + upstream + correlated faults NOT covered; N-version is the stronger step).
  - §5 what Option B (N-version) would add — honest about cost.
  - §6 test surface — AGREEMENT input count, DETECTION fault-injection result.
  - §8 open items for review (proptest scaling, N-version, process/channel isolation, audit-sink wiring, WCET impact).

Traceability matrix regenerated.

Verified locally:
  - cargo build --workspace (root) + cd parko && cargo build --workspace → clean
  - cd parko && cargo test -p parko-kirra → 61 lib + 3 integration passed (was 57+3; +4 diversity_tests with 2 fault-injection detection tests + diverse::tests' 5)
  - cd parko && cargo test --workspace --exclude parko-onnx --exclude parko-openvino → 186 passed, 0 failed, 0 ignored (was 177 → +9 net for the L2 work)
  - cargo test --workspace (root) → 636 passed, 0 failed, 11 ignored
  - cargo test -p kirra-runtime-sdk --lib traceability_gate → 4 passed

What still needs human safety-engineer review (per the doc §8):
  - Coverage adequacy of the 30-input AGREEMENT sweep (proptest follow-up).
  - N-version reimplementation (Option B) if spec-level systematic- fault coverage becomes a target.
  - Process / physical channel isolation (correlated environmental faults not addressed at the governor layer).
  - Production audit-sink wiring (InMemoryDivergenceSink is test-only; production must persist to AuditChainLinker).
  - WCET regression check now that the shadow is doing independent work (was identical-cost before; now doubles the per-tick computation as expected by design).


justinlooney:
Closing as **superseded** — the CERT-006 diverse-shadow GovernorComparator already landed on `main` via a separate branch before this PR could merge:

- `63b407b` — `feat(parko-kirra): diverse second governor for the comparator — CERT-006` (the equivalent feature: `diverse.rs`, `COMPARATOR_DIVERSITY.md`, and the `comparator.rs`/`lib.rs` wiring).
- Plus follow-up safety fixes that this branch predates and does **not** contain:
  - `29a785d` — reachable + mutation-verified diverse ODD-cap (finding 2)
  - `3782a62` — Degraded posture = decel-to-stop-and-HOLD (not an MRC crawl)
  - `10a2618` — quality-hardening property/fault-injection tests

The on-`main` implementation is the more complete one (`diverse.rs` ~810 lines vs this branch's ~483) and carries the later safety fixes. Because both branches created `diverse.rs` and `docs/safety/COMPARATOR_DIVERSITY.md` from scratch, the merge conflicts add/add; resolving in this branch's favor would regress the safety fixes above, and resolving in `main`'s favor would make this merge a functional no-op.

No code changes are being made. Closing in favor of the version already on `main`. If there's a specific improvement in this branch's implementation worth porting over, it's best done as a small targeted PR against the current `main` rather than merging this parallel reimplementation.

---

PR #777, opened 2026-07-03
kpi: scenario-KPI CI gate — unsafe_miss_rate / admissibility / hazard_recall over a generated corpus (WS-3.1)

justinlooney:
**PR 8 — the first post-GATE-A item: WS-3 Phase 1 (scenario & evidence engine), closing the Stage-1 slice of G9.**

## What it is

A new `crates/kirra-kpi-gate` crate + CI job that thresholds the three fleet-safety KPIs the plan names — **`unsafe_miss_rate`**, **admissibility**, **`hazard_recall`** — over a parameterized scenario corpus in the low hundreds. Red = a fleet-safety KPI regressed past the reviewed policy; the WS-3 DoD ("no safety-relevant PR merges without the KPI gate") starts here.

Three design properties:

- **It consumes the existing metric harnesses, never re-implements one.** Admissibility runs every proposal through the *real checker* (`validate_trajectory_slow` via `kirra_doer_eval::verdict_of`) into the existing `AdmissibilityTally` (fail-closed: an empty corpus scores 0.0). Perception scores through `kirra_taj::SemanticEvalSummary` at the fusion oracle. Two doers are scored: the shipped `GeometricPlanner` and the seeded SafetyAware `LearnedPlanner` (the quantization-scorecard seam).
- **Deterministic.** Closed-form generators (no RNG, no time), seeded planner — a red gate is a real behavior change, never flake. Corpus sizes are **pinned by test** (248 doer scenarios: hazard kind {stopped, lead-moving, oncoming} × 10 distances × 4 ego speeds × 2 goals + clear-road rows; 71 perception frames: class {Water, StaticObstacle} × 11 distances × 3 lateral spans + clear frames) — a silent shrink can't weaken the gate while reporting green.
- **The bar predates the model** (the taj crate's own discipline): today's perception frames run through the scripted `MockSemanticDetector` **seam** and pin perfection; when the real RGB→TensorRT detector lands behind the same trait, it inherits corpus and thresholds on day one.

## The policy

`ci/scenario_kpi_thresholds.json` — bounds set **at the measured floor** (exact rationals, since everything is deterministic): geometric admissibility 240/248 = 0.9677, learned 224/248 = 0.9032, `unsafe_miss_rate` ≤ 0.0, `hazard_recall` ≥ 1.0. Moving a bound down requires a justified change to the file in the PR; improvements tighten it (ratchet). `deny_unknown_fields` so a typo'd threshold can't silently default.

## DoD proven both ways

- **Green:** the gate passes against the committed thresholds — also pinned as a unit test, so plain `cargo test` catches a KPI regression even without the CI job.
- **Red:** `gate_goes_red_on_a_kpi_breach` (unreachable bound → red, attributed to the right row); `blind_detector_breaches_the_perception_kpis` (a detector that sees nothing: `unsafe_miss_rate` > 0.5, recall 0 — the metric is live, not vacuously green); binary-level runs (breach → exit 1 with the failing row named; missing/unparsable policy → exit 2).

Supporting changes: `EvalScenario.name` `&'static str` → `String` (generated names) and read-only world accessors on `EvalScenario`; the new CI job runs `cargo run --locked -p kirra-kpi-gate` on every push/PR.

118 workspace suites green; clippy zero warnings; guardrails + deny gates pass.

https://claude.ai/code/session_01MC8UD3ajLiTY7QtnNE7oJ6

---
_Generated by [Claude Code](https://claude.ai/code/session_01MC8UD3ajLiTY7QtnNE7oJ6)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,22 @@ jobs:
           command: check advisories
           arguments: --locked
 
+  # WS-3.1 — the scenario-KPI gate: thresholds unsafe_miss_rate /
+  # admissibility / hazard_recall over the generated deterministic corpus
+  # (248 doer scenarios + 71 perception frames) against the reviewed policy
+  # in ci/scenario_kpi_thresholds.json. Red = a fleet-safety KPI regressed
+  # (the corpus and planners are deterministic — never flake). The WS-3 DoD:
+  # no safety-relevant PR merges without this gate.
+  scenario-kpi-gate:
+    name: scenario-KPI gate (WS-3.1)
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    steps:
+      - uses: actions/checkout@v7
+      - uses: dtolnay/rust-toolchain@stable
+      - name: KPI gate (unsafe_miss_rate / admissibility / hazard_recall)
+        run: cargo run --locked -p kirra-kpi-gate --bin scenario_kpi_gate
+
   test:
     name: Test
     runs-on: ubuntu-latest
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2126,6 +2126,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "kirra-kpi-gate"
+version = "0.1.0"
+dependencies = [
+ "kirra-core",
+ "kirra-doer-eval",
+ "kirra-planner",
+ "kirra-taj",
+ "kirra-trajectory",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "kirra-l3-e2e"
 version = "0.1.0"
 dependencies = [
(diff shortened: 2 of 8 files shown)


justinlooney:
Re the Copilot comment on `EvalScenario.name: &'static str → String`: no change needed — the claimed compile error doesn't occur.

All four `sc.name` usages in `crates/kirra-doer-eval/tests/onnx_roundtrip.rs` are `assert!`/`assert_eq!` **format arguments**, and `format_args!` takes its arguments by reference implicitly — accessing a `String` field through `&EvalScenario` in a format position borrows it, it doesn't move it. A move-from-borrow error would require something like `let owned: String = sc.name;`, which no caller does.

Verified explicitly: `cargo test -p kirra-doer-eval --test onnx_roundtrip --locked` compiles and passes (4/4), as did the full 118-suite workspace run before push.

---

PR #946, opened 2026-07-15
Add clean-room ROSMASTER R2 MCU firmware foundation

justinlooney:
## Summary
- add an evidence-ranked, non-enabling clean-room hardware baseline for the Yahboom ROSMASTER R2 and quarantine unresolved PCB/harness facts
- add an allocation-free C++17 portable core covering HAL contracts, bounded Ackermann kinematics, steering-consistent wheel control, anti-windup PID, fail-closed safety transitions, R2CP framing, diagnostics, transactional page-aligned configuration rollback, and authenticated-image seams
- add architecture, protocol, safety, calibration, manufacturing, developer, and Phase 0–10 production roadmap documentation
- add host CMake tests, deterministic simulation, sanitizers, pinned clang-tidy CI, cppcheck, Doxygen, coverage, and SHA-pinned actions

## Scope
This is the portable Phase-1 foundation, not a flashable board image. The unresolved R2 harness manifest keeps actuation disabled. STM32 BSP/RTOS, reviewed crypto, signed bootloader backend, Linux ros2_control bridge, physical calibration, HIL, and target WCET/latency evidence remain explicit phase gates.

## Safety decisions
- no Yahboom firmware source or X3/mecanum architecture is inherited
- POST/fault states cannot transition directly into an armable state
- controlled stop retains only internal deceleration authority until measured stop
- invalid/overflowing geometry, limits, gains, measurements, message IDs, flags, sequences, or configuration fail closed
- Kirra command authorization is distinct from R2CP link authentication
- the stock 115200-baud link cannot meet the sub-2 ms end-to-end target for the ~70-byte command frame
- STM32F103 software verification is not represented as hardware-backed secure boot or tamper-resistant rollback

## Verification
- dedicated workflow run 29439968268: build, sanitizer tests, deterministic simulation, clang-tidy 18, cppcheck, Doxygen, and coverage all pass
- local line coverage: 88.0% (629/715); function coverage: 90.9% (70/77)
- independent embedded-core and architecture reviews report no remaining portable Phase-1 merge blockers

[rosmaster_r2_firmware_final_verification_2.log](https://cursor.com/agents/bc-50b9654e-ba29-4604-a652-c1e82868d1a8/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Frosmaster_r2_firmware_final_verification_2.log)

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-50b9654e-ba29-4604-a652-c1e82868d1a8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-50b9654e-ba29-4604-a652-c1e82868d1a8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


firmware/rosmaster-r2/safety/src/safety_manager.cpp
```diff
@@ -0,0 +1,182 @@
...
+    constexpr auto immediate_disable_faults =
+        bit(Fault::emergency_stop) |
+        bit(Fault::battery_overcurrent) |
+        bit(Fault::motor_overtemperature) |
+        bit(Fault::control_deadline_missed) |
+        bit(Fault::encoder_implausible) |
+        bit(Fault::motor_runaway) |
+        bit(Fault::steering_implausible) |
+        bit(Fault::self_test_failed);
+    return state_ == SafetyState::boot ||
```
justinlooney: **HIGH — `bridge_must_be_disabled()` fails *open* on an unknown/corrupt state (wrong polarity).**

This predicate is an allowlist of *safe* states, so any `state_` value not in the set — with no immediate-disable fault bit — returns `false` → **bridge stays enabled**. That's default-*allow* for a hardware-enable gate, the opposite polarity of the (correct) deny-by-default `motion_permitted()`. `state_` is a `std::uint8_t`; an SEU / wild write to an out-of-range value (e.g. `0x40`) with `active_faults_ == 0` yields bridge-enabled with no valid command source.

Fix: invert to an allowlist-of-*actuating* states — "disabled unless `state_ ∈ {active, controlled_stop}` with a clean immediate-disable set" — so an unknown state disables the bridge. This is the single most important pre-actuation fix and it's a small change.

---

firmware/rosmaster-r2/safety/src/safety_manager.cpp
```diff
@@ -0,0 +1,182 @@
...
+void SafetyManager::evaluate(const SafetyInputs& inputs) noexcept {
+    active_faults_ = 0U;
+    if (inputs.emergency_stop_asserted) {
+        raise(Fault::emergency_stop, true);
+    }
+    if (!inputs.command_fresh &&
+        (state_ == SafetyState::active || state_ == SafetyState::controlled_stop)) {
+        raise(Fault::command_timeout, false);
+    }
+    if (!inputs.battery_voltage_safe) {
```
justinlooney, line 77: **HIGH — battery undervoltage (and `command_timeout`/`imu`/`comms`) are non-latched with no hysteresis → chatter.**

`evaluate` clears `active_faults_` every cycle, and these are raised with `latch=false`. Voltage sag-under-load is the canonical battery failure mode: ACTIVE → motors load the pack → undervoltage → `controlled_stop` → motors stop → voltage recovers → `motion_stopped` → `standby` → re-`arm`/`activate` → sag again → **oscillates safe↔armed indefinitely**, and never gate-blocks re-arm once the reading momentarily clears.

A safety-relevant fault should require an explicit recovery step (latch) or at minimum an N-of-M debounce + recovery hysteresis band. Recommend latch-on-confirm for battery/thermal-adjacent channels and a debounce counter on the rest.

Related (declared-but-dead): `Fault::brownout` and `Fault::watchdog_precursor` are never raised anywhere, and `SafetyInputs` has no brownout/watchdog fields — so the PR's "fail-closed on brownout/watchdog" invariant isn't enforced in this layer. Either wire real inputs + `evaluate` branches, or drop the claim.

---

firmware/rosmaster-r2/bootloader/include/r2/boot/image_verifier.hpp
```diff
@@ -0,0 +1,45 @@
...
+struct ImageManifest {
+    std::uint32_t product_id;
+    std::uint32_t hardware_compatibility;
+    std::uint32_t image_length;
+    std::uint32_t security_version;
+    std::uint8_t digest_sha256[32];
+    std::uint8_t signature_ed25519[64];
+};
+
+enum class VerificationResult : std::uint8_t {
```
justinlooney, line 22: **HIGH — `VerificationResult::accepted` is the zero value → fail-open default.**

A default-constructed / zero-initialized / `memset` verdict reads as `accepted`. Any stub or partially-implemented verifier, or a `VerificationResult r{}; … use(r)` path that forgets to assign on an error branch, **accepts an unauthenticated image**. A security verdict's `0` must be the reject state. One-line reorder: put `rejected`/`unknown` first.

Two companion notes on this header:
- The seam advertises `digest_sha256[32]`, `signature_ed25519[64]`, and `rollback_attempt`, which reads as authenticated secure boot — but the header has **zero comments**, and the PR's own caveat ("STM32F103 software verification is not hardware-backed secure boot") appears nowhere an integrator would read it. Add a header comment: software-integrity only, no hardware root of trust on the F103.
- Confirm the (not-in-this-PR) `verify()` implementation rejects on *every* mismatch branch and returns the reject state by default.

---

firmware/rosmaster-r2/protocol/src/wire.cpp
```diff
@@ -0,0 +1,303 @@
...
+        return DecodeStatus::unsupported_version;
+    }
+
+    const auto payload_length = get_u16(&decoded[6]);
+    if (payload_length > kMaximumPayload ||
+        decoded_length != kHeaderSize + payload_length + kCrcSize) {
+        return DecodeStatus::invalid_length;
+    }
+    const auto body_length = kHeaderSize + payload_length;
+    if (get_u32(&decoded[body_length]) != crc32c(decoded.data(), body_length)) {
```
justinlooney, line 249: **HIGH (architectural, disclosed) — CRC-32C is integrity, not authentication.**

Credit where due: the decode path is genuinely fuzz-grade — every wire-driven length/offset is bounds-checked before use (the exact-length equality at the `decoded_length != kHeaderSize + payload_length + kCrcSize` check is the load-bearing line), CRC is verify-before-use, and unknown version/type/flags all fail closed. I could not construct an OOB read.

But CRC-32C is the *sole* gate, and it's non-cryptographic: anyone who can put bytes on the SBC link computes a valid CRC and forges a fully-legitimate frame — including `enter_bootloader` / `firmware_block` / `firmware_commit`. If the link is trusted/physical this is fine (and the docs correctly defer the MAC to a phase gate); flagging because the brief frames the link as untrusted. Prioritize a MAC/signature over the `firmware_*` and arm/activate/motion frames if that threat is in scope.

Two smaller protocol items: `SequenceTracker` is trust-on-first-use **and `decode()` never calls it** — a caller that forgets `accept()` gets zero replay protection, so either bind it into the accepted-frame path or document the MUST. And please add the libFuzzer/`FuzzedDataProvider` decode target that `docs/PROTOCOL.md` already mandates — the exact-length `!=` check is safety-critical and deserves a regression/property test so a future change to `<` can't silently reintroduce an OOB window.

---

justinlooney (commented):
## Architect-level review — clean-room ROSMASTER R2 MCU firmware foundation

Reviewed the branch across five dimensions (safety state machine, motion/kinematics numerics, R2CP wire protocol, flash-config + boot-image seam, and build/CI/tests/docs), reading the actual sources. Inline comments are attached to the four most surgical HIGH findings.

### Verdict

**A genuinely serious, well-engineered Phase-1 foundation — and refreshingly honest in its docs (it under-claims if anything).** As a *portable, non-actuating* baseline it's close to mergeable-as-draft. It is **not** yet a foundation that could gate real actuation: a handful of HIGH issues (one a real safety-polarity bug), plus a testing gap on the one attacker-reachable parser that the 88% line-coverage headline hides.

### What genuinely holds up (most of the PR)

- **Clean-room provenance is real and well-quarantined** — `HARDWARE_REFERENCE.md` is evidence-ranked (`Evidence` enum mirrored in `board_manifest.hpp`), pins Yahboom repos as *sources*, quarantines the ICM20948-vs-MPU9250 IMU conflict, and refuses the battery divider as a default. Nothing reads as inherited firmware. The core brief is satisfied.
- **The config store is genuinely power-loss safe** — CRC'd A/B ping-pong with generation-as-atomic-commit; the crash cases were walked and there's no torn-config window (bad-CRC → reject → fall back to intact slot / factory defaults, fail-closed).
- **The R2CP decode path is memory-safe against arbitrary bytes** — every wire-driven length/offset is bounds-checked before use, CRC is verify-before-use, unknown version/type/flags fail closed. No OOB read exists.
- **The motion math is correct and fail-safe** — correct rear-drive bicycle geometry, no reachable div-by-zero/singularity, clamp-before-wheel-derivation, envelope-before-rate ordering, correct PID anti-windup with `dt` as a parameter; no non-finite path reaches a command.
- **The safety core algebra is fail-closed** — safe default state, deny-by-default `motion_permitted()`, no fault→armed edge, controlled-stop holds internal-decel-only until measured stop.
- **Infra fundamentals** — `.clang-tidy` is real and gating (and doesn't falsely claim MISRA), actions are SHA-pinned, firmware CI is path-isolated from the Rust workspace, and the unit tests that exist assert real invariants (power-loss injection, NaN guards, latch-clear-requires-ack).

### HIGH (see inline for H1, H2, H4, H5)

- **H1 — `safety_manager.cpp` `bridge_must_be_disabled()` fails *open* on a corrupt/unknown state** (wrong polarity vs deny-by-default `motion_permitted()`). *Real safety bug; small fix.* [inline]
- **H2 — battery-undervoltage & `command_timeout`/`imu`/`comms` are non-latched, no hysteresis → chatter safe↔armed under load.** [inline]
- **H3 — `Fault::brownout` / `watchdog_precursor` are declared but never raised** (no `SafetyInputs` fields) — the "fail-closed on brownout/watchdog" invariant is declared-but-dead in this layer. Wire it or drop the claim.
- **H4 — `image_verifier.hpp` `VerificationResult::accepted == 0` → fail-open default**; plus the crypto-implying interface has no "software-integrity-only, no hardware RoT" disclaimer in-code. [inline]
- **H5 — CRC-32C is integrity, not authentication** → forged frames (incl. `enter_bootloader`/`firmware_*`) accepted on an untrusted link. Disclosed in docs; flagged because the brief frames the link as untrusted. [inline]
- **H6 — the decoder is materially under-tested** — 9 rejection branches, exactly one negative decode test, and **no fuzzer** despite `PROTOCOL.md` mandating one. This is what the line-coverage number hides.
- **H7 — UBSAN is non-halting** (`-DR2_ENABLE_SANITIZERS=ON` but no `-fno-sanitize-recover` / `UBSAN_OPTIONS=halt_on_error=1`) → a UBSAN finding logs but `ctest` still exits 0. Decorative as configured.

### MEDIUM

- `configuration.cpp` — a `!calibrated` record skips **all** range checks, so a CRC-valid uncalibrated record can carry a `NaN`/`0`/huge gain; safety rests on every consumer immobilising when `!calibrated`.
- `motion_controller.cpp` — a sharp-turn-from-standstill can trip `infeasible_stationary_turn` on the rate-limited speed and hard-`reset()` mid-maneuver (steering re-centered) → stutter/stall. Direction is safe; maneuver quality isn't.
- `safety_manager.cpp:107-110` — `controlled_stop→standby` gates only on `motion_stopped`, ignoring still-active non-latched faults (publishes "healthy" under a live fault).
- **Coverage gate is line-only 80%** — no branch/MC-DC despite `SAFETY_AND_PRODUCTION.md` invoking MC/DC; most of the fault-policy table (overcurrent/thermal/encoder/steering/runaway/deadline) is never individually injected.
- **The actuation gate is doc-not-code** — `kUnresolvedR2Harness{…actuation_enabled=false}` is never referenced by a `static_assert`/BSP; add a `static_assert(!kUnresolvedR2Harness.actuation_enabled)` in any future actuation TU so the gate is compiler-enforced.

### LOW

One-tick input glitch `reset()`s steering to center instead of holding last-good; no odometry/heading-wrap/covariance in these files; `wheel_radius_m` validated but unused; `SafetyInputs` is a 12-field positional-bool aggregate (use designated initializers); `check.sh` skips tidy/cppcheck if tools absent; unused `kConfigurationSlotBytes`.

---

## Schema gaps surfaced by mapping the existing bench data onto `PlatformConfiguration`

We have prior **bench-measured R2 facts** (`docs/hardware/HARDWARE_FINDINGS_R2X3.md`, `R2_PATH_B_ACKERMANN_DRIVE.md`, `robot/r2_drive_calibration_results.txt`). Three drop straight into this firmware — `wheel_radius_m = 0.0333375` (66.675 mm caliper diameter, retires the unused-field LOW), `left_encoder_counts_per_revolution = 834`, and the `M1=rear-left / M4=rear-right, both +forward` map (already reflected in `kUnresolvedR2Harness`). Mapping the rest exposed four schema gaps worth addressing:

- **G1 — `uint32 counts_per_revolution` can't hold the measured `834.5`** (0.06% odometry scale error; single 2-turn sample). Consider a higher-resolution scale field + a multi-sample-mean requirement.
- **G2 — there is no PWM↔speed / motor-output scale field.** The real drive calibration (`V_PER_PWM ≈ 0.0145 (m/s)/PWM`, offset, deadband, **per wheel**) has nowhere to live, so the firmware can't honour a physical `m/s` command without hard-coding it in the BSP. Add per-wheel `mps_per_duty` (slope + deadband) fields, or an explicit closed-loop-only contract. **Most actionable gap.**
- **G3 — no motor-direction / encoder-sign field** in `board_manifest` for the "both forward = +" and per-encoder counting-sign HAL facts.
- **G4 — a measured ~28% left/right tick imbalance at equal PWM is unresolved** (real drift vs encoder-scale) → "equal command ⇒ straight" is not guaranteed; also bears on odometry covariance. One capture (RR ticks/rev) closes it.

Also note: the earlier steering data (`[-45,+45]` command units, centre trim) is **Yahboom-`Rosmaster_Lib`-protocol**, not raw µs — this PR's direct TIM7 servo path needs a fresh raw-PWM sweep to fill `servo_*_us`; the existing data cannot. And the bench's "negative = left, sign STILL to verify" must be reconciled against this firmware's `ω>0 → δ>0 → left` convention before any turn is enabled.

## Recommendation

For a **draft Phase-1 portable foundation** this is strong work; I'd support merging it as such after the cheap fail-closed fixes — **H1** (default-deny the bridge predicate), **H4** (reorder so reject == 0), **H3** (wire or drop brownout/watchdog). Before it can **gate real actuation**: H2 (latch/hysteresis), H6+H7 (real decoder fuzzer + halting UBSAN), the `!calibrated` validation gap, and — if the SBC link is in the threat model — H5 (a MAC over `firmware_*`/control frames). The docs already scope most of these as phase gates, which is the right posture; the remaining move is to make two of them (the actuation gate, the fuzzer) **load-bearing in code/CI** rather than in prose.

---

PR #980, opened 2026-07-16
refactor(kirra-verifier-pg): split PgVerifierStore seams into per-store modules

justinlooney:
## Summary

`kirra-verifier-pg/src/lib.rs` was 1534 lines: `PgVerifierStore` with **22 impl blocks** (one trait impl per storage seam, several with a small inherent row-mapping helper impl + a `*_COLUMNS` const), all in the crate root (the #3 monolith from the de-monolith review). Move each seam into its own submodule — mirroring the per-table layout `kirra-persistence` uses on the SQLite side it conforms against.

## The split

| File | Lines | Seam |
|---|---|---|
| `lib.rs` | 475 | schema/migration engine (`BASELINE_DDL`, `PG_MIGRATIONS`, `PG_SCHEMA_VERSION`), `LivePgExecutor`, `PgStoreError`, `PgVerifierStore` struct + core inherent impl (`connect`/`from_client`/`initialize`/`schema_version`/`lock`/`row_to_node`) |
| `nodes.rs` | 68 | `NodeStore` |
| `epoch.rs` | 75 | `EpochFence` |
| `posture.rs` | 84 | `PostureEngineStateStore` |
| `federation.rs` | 109 | `FederationStore` |
| `operators.rs` | 79 | `OperatorStore` |
| `principals.rs` | 90 | `PrincipalStore` |
| `cert_principals.rs` | 112 | `CertPrincipalStore` |
| `fabric.rs` | 88 | `FabricAssetStore` |
| `ota_campaigns.rs` | 235 | `OtaCampaignStore` |
| `av_subsystem.rs` | 190 | `AvSubsystemStore` |

`lib.rs` **1534 → 475 lines**.

## Behavior-preserving — pure relocation

- The only visibility change is `pub(crate)` on the two shared helpers `lock` and `row_to_node` so the split impl blocks reach them. Each store's own row-mapping helpers move with it and stay private.
- **No public API change** — all trait impls remain on `PgVerifierStore` at the same paths.
- Verified in the workspace-detached crate (the `postgres-conformance` CI lane's `cd crates/kirra-verifier-pg && cargo …`): **lib + `tests/live_pg.rs` compile**, clippy clean, fmt clean. (The live conformance tests skip without `KIRRA_PG_URL`; the CI lane provides a `services: postgres` container.)

De-monolith #3, after #978 (GeometricPlanner) and #979 (LaneGraph). Remaining candidates: `kirra-ota-ctl.rs` (CLI subcommands), `standby_monitor.rs`.

https://claude.ai/code/session_01JoequnLBhG4EcNywrMsBmr

---
_Generated by [Claude Code](https://claude.ai/code/session_01JoequnLBhG4EcNywrMsBmr)_


justinlooney:
Thanks — grouping the 5 Copilot comments (all the same `u64`↔`i64` timestamp-cast point across nodes/operators/principals/epoch/`row_to_node`).

Leaving them as-is in **this** PR, deliberately:

1. **Pre-existing, moved verbatim.** These casts were on `main` before this change (e.g. `registered_at_ms: row.get::<_, i64>(2) as u64`); this PR is a pure relocation of `PgVerifierStore`'s seam impls into per-store modules with no behaviour change. Folding a semantic fix into it would make the refactor harder to review/revert.

2. **Changing only PG would break backend parity.** `kirra-verifier-pg` is conformance-tested for byte-parity against the SQLite backend, and the SQLite side uses the *same* `as u64`/`as i64` pattern for these timestamps (`kirra-persistence/src/nodes.rs`). So the premise "other read paths treat negative BIGINT as 0" doesn't hold for these node/operator/principal timestamps — both backends cast. A checked-conversion fix is worth doing, but it has to land on **both** backends together (and against the shared conformance suite), not on the relocated PG lines alone.

Tracking it as a separate follow-up (cross-backend timestamp-cast discipline) rather than expanding this relocation.

---

PR #1186, opened 2026-07-27
feat(r2): evidence-bound V2 actuation chain, end to end

justinlooney:
> **Upgrade note:** Existing `/etc/kirra/robot.env` files are preserved by the installer. Operators must manually add `KIRRA_PROFILE_DIGEST` before restarting the V2 consumer.

Binds every actuator command to the perception evidence it was authored against, from the doer through to the motor.

Occy/Taj evidence → atomic ROS proposal envelope → interceptor `release_binding` → verifier V2 mint → 176-byte payload + 96-byte token → robot V2 FFI consumer → single motor-write chokepoint.

## Why

A bare `Twist` names no evidence, so the governor could only ever sign "some robot asked for 0.4 m/s". Splitting command from evidence across topics also let them go stale independently. V2 makes them one atomic message and binds the perception/profile/proposal identity inside the signed bytes, so the motor consumer can refuse a release minted for a superseded frame or another platform.

## Fail-closed properties

- A malformed, empty, or unbound proposal stops the robot and makes no request.
- No previous Twist or binding is ever reused.
- A 128-byte V1 frame is refused at both the relay and the consumer — no silent downgrade.
- Refusals leave all watermarks unchanged and never feed liveness, so a flood starves into the SS-002 stop exactly as silence does.
- The pinned profile digest comes from trusted configuration only; unset or malformed aborts startup, mismatched refuses every release.

## V1 is retained

V1 stays in the Rust library, `kirra_ffi.KirraConsumer`, and the minter's `frame` subcommand, still covered by the V1 half of the smoke test. It is simply not on the live R2 path.

## Tests

Extensive new coverage across Rust, ROS Python, robot FFI, and operator diagnostics: the envelope grammar and interceptor callback (no ROS graph), the V2 wire parse and serial chokepoint (no ROS/motor/`.so`), and the FFI boundary against the real `.so` with real governor-signed frames — a positive release plus every refusal arm, each reached by breaking exactly one field. All wired into the existing CI steps.

## Operators

`KIRRA_PROFILE_DIGEST` is a new required variable. Already-provisioned robots must add it by hand, since the installer never overwrites an existing `robot.env`. Obtain and verify it with:

```bash
python3 robot/kirra_doctor.py --module profile_digest   # lidar + Taj up
```

The same value must be exported for the bench publisher; a mismatch refuses every release with `PROFILE_DIGEST_MISMATCH`. A fresh install renders the pin as a `__FILLED_...__` placeholder — the consumer refuses to start and both the `config` and `governor` doctor modules FAIL on it until a real value is pinned.

## Validation

`cargo fmt --check`, `clippy -D warnings` (release-token all-targets + consumer crates), 110 Rust crate tests, `actuator_envelope_integration` (19), `ros_release_mint_tests` (6), `gateway::policy_layer` (13), ROS package pytest (190), robot suites (62 under pytest; all four also exit 0 under their native runners), plus `py_compile` and `bash -n` across the changed Python and shell.

## Notes for reviewers

- Includes the 7 prerequisite V2 commits (`5152c73`…`78f0a62`), which were never merged to `main`.
- `kirra_with_robot.launch.py` still wires nav2 into `/cmd_vel_raw`, now a `String` topic. Occy is the live doer, so no live path breaks, but that launch option is now inert.

---


justinlooney:
## CI status

**Fixed and pushed** (`a945299e`) — *Static Analysis (ISO 26262)* failed the quality guardrail:

```
- src/gateway/policy_layer.rs: 1621 lines > guardrail max 1526
```

`main` sits exactly at the 1526 ceiling and this branch adds 95 lines. The growth is the actuator release-binding perimeter from `4c19786`: `ActuatorReleaseBinding`, `decode_lowercase_sha256`, `ValidatedActuatorReleaseBinding`, `validate()`, the `ActuatorMotionCommandRequest` wrapper, and threading the binding through the three clamp re-serialization arms.

Per the gate's own instruction I raised the ceiling to exactly 1621 rather than refactoring reviewed code late in the PR, with the justification recorded in the baseline's `_comment`. The placement is deliberate: untrusted binding metadata is validated **at the HTTP perimeter**, where it arrives, and converted to fixed-width digest bytes so no later code can sign malformed, uppercase or truncated text — and the reusable `ProposedVehicleCommand` kinematics kernel stays free of it. Panic budget is unchanged at 14; every new path returns `Option` or `StatusCode::BAD_REQUEST`, none unwrap. The ceiling is re-pinned exactly, so the next growth still fails.

**Not actionable here** — the seven `Vercel – kirra-runtime-sdk-*` statuses are red with *"Deployment rate limited — retry in 24 hours."* That is an account-level build quota, not a build error: this PR touches no `website/` or `console/` files, so there is nothing in the diff for a preview deploy to break. They should clear on their own once the window resets. Flagging so nobody spends time chasing them.

The rest of CI is re-running on `a945299e`. I'm watching and will push fixes for anything real.

---

Issue #1211, opened 2026-07-28
[R2 perception 2/8] R2 sensor watchdogs — a frozen LiDAR is currently indistinguishable from a stationary world


justinlooney:
Child of #1209. Audit phase 7. **Status: CLOSED — scan liveness is DEPLOYED, not merely implemented.**

> Landed in #1222 (checker) and #1223 (live wiring). A frozen `/scan` now floors the served speed cap on the real robot. Verified end to end against the running service binary, not only in tests: a healthy stream serves 4.099 m/s; a driver replaying one buffer drops it to 0.000 with `SENSOR_NO_MESSAGES` and `healthy=false`.

## The failure this closed

Generic watchdogs detect **silence** — `telemetry_watchdog.rs`, the armed-but-silent channel resolvers, `KIRRA_SUBSCRIPTION_STALENESS_MS`. A LiDAR driver that keeps republishing its last frame, or one that died behind a latched publisher, answers every one of them satisfactorily. The topic was fresh by timestamp and silent by no metric; the world looked stationary and clear; the checker admitted motion into a scene nobody had observed since the driver stalled.

Freshness is not liveness. That distinction is now enforced.

## Delivered

**Checker (#1222)** — `kirra_trajectory::sensor_watchdog`, pure, no ROS.

- [x] **Publisher count** — `SENSOR_NO_PUBLISHER`, distinct from silence (different operator fix: wiring vs restart)
- [x] **Frozen timestamp** — `SENSOR_FROZEN_TIMESTAMP`; a stamp going *backwards* counts too (clock jump / replayed bag)
- [x] **Frame fingerprint freeze** — `SENSOR_FROZEN_FRAME` over a bit-exact `fingerprint_from_ranges`. The separation from a genuinely static scene comes from **bit-exactness, not the repeat count**: a real LiDAR moves the low bits of nearly every return, so two bit-identical frames are the same buffer twice. The conservative residual (an all-no-return scan trips it) is accepted deliberately and documented — such a scan carries no evidence of clear space either.
- [x] **Valid-ray ratio** — `SENSOR_LOW_VALID_RAY_RATIO`, including the zero-ray frame
- [x] **Measured frequency** — `SENSOR_RATE_BELOW_FLOOR` carrying `measured_hz` and `floor_hz`, a number not a boolean
- [x] Distinct reason codes, walked by a test so a new variant reusing one reds
[...]

Issue #1210, opened 2026-07-28
[R2 perception 1/8] LiDAR self-filter calibration — robot returns are not identified or masked


justinlooney:
Child of #1209. Audit phase 6. **Verified status: gap — zero implementation.**

> **Progress (branch `claude/ros-bound-proposal-migration-nso5d2`, commit `797f6ae0`).** The machinery has landed: `kirra_taj::self_filter`, wired into `ingest_scan` — the single funnel every downstream stage consumes — plus per-reason rejection counting on the perception wire. 45 tests. **Default is disarmed and the pipeline is behaviourally unchanged**, because the physical capture that says *which returns are the robot* still needs the hardware. Procedure written up in `docs/hardware/R2_SELF_FILTER_CALIBRATION.md`. Remaining boxes are the hardware ones.

## Verified current state

A search for self-filter artifacts across `crates/`, `ros2_ws/`, `robot/` and `parko/` returns nothing: no mask polygons, no laser-to-base transform verification, no mask-boundary tests, no rejected-point metrics. This is the most complete absence in the plan.

The lateral-return eligibility fix that already landed is **not** equivalent. That fix stops a lateral return from being treated as a frontal collision object; it does not establish which returns are the robot itself.

## Why this is first, ahead of items the plan ranked higher

Until self-filtering is calibrated, the robot may be perceiving parts of itself. Those returns enter tracking, fusion and the checker as ordinary world geometry, which silently corrupts every downstream phase that consumes tracks — including the ones the audit confirmed as substantially complete. A phantom self-return near the body reads as a stationary obstacle at zero range, and the correct fail-closed response to that is to stop.

## The asymmetry that shapes the design

Worth stating because it inverts the usual rule. Everything else in Taj **tightens** — an object narrows the corridor, a bad scan drops confidence, a hazard clips the clear distance — so a bug in any of them fails toward a stop. Discarding a return does the opposite: the corridor widens, an object disappears, the assured-clear distance grows. A mask one polygon too generous does not make the robot cautious; it makes it blind to whatever is standing in the region it was told to ignore.

Every constraint in the landed code exists to bound how much evidence a mask can destroy.

## Acceptance criteria

**Software — landed**

- [x] Filter applied **before tracking**, at `ingest_scan`, with a per-reason rejected-point counter (`not_a_number`, `below_min_range`, `beyond_max_range`, `self_masked`) riding `TajPerception` to `PerceptionResponse.rejected`.
[...]

PR #1221, opened 2026-07-28
feat(taj): add bounded self-filtering with verified mount provenance

justinlooney:
> Unlike other Taj processing, self-filtering can widen the corridor by deleting returns. Therefore every mask constraint is designed to bound evidence destruction, and the feature ships disarmed until hardware calibration exists.

Closes the software half of #1210. The hardware calibration is documented, not performed — see *What is deliberately not here*.

## The asymmetry

Every other operation in Taj **tightens**. A detected object narrows the corridor, a bad scan drops confidence, a semantic hazard clips the clear distance. A bug in any of them fails toward a stop.

Self-filtering is the only operation that goes the other way. Discarding a return **adds** free space: the corridor widens, an object disappears, the assured-clear distance grows. A mask one polygon too generous does not make the robot cautious — it makes it blind to whatever is standing in the region it was told to ignore.

That inversion is why the constraints below are shaped the way they are, and why the default is off.

## What reviewers should check

Seven claims carry the safety argument. Each has a test; the parenthetical is where to look.

**1. Default remains disarmed and behaviourally identical.**
`TajPhaseA::self_filter` is `None` unless explicitly constructed otherwise. No mask, no transform, no filtering. (`the_default_pipeline_is_untouched` asserts an unarmed pipeline is byte-identical to the pre-change path; the existing 140-test Taj suite is unmodified and green.)

**2. Mask geometry is rejected, not silently clamped.**
`SelfFilterMask::new` returns `Err(SelfFilterError::ExceedsRadius { .. })` naming the offending vertex. A clamp would be worse than a refusal: the operator would believe the robot filters a region it does not, and the difference only shows up as an obstacle nobody saw. (`an_oversized_mask_is_refused_not_clamped`.)

**3. Boundary points remain as evidence.**
A return exactly on a polygon edge or vertex is **retained**. Ties break toward keeping evidence, consistently across every edge and both winding orders. (`a_return_exactly_on_the_boundary_is_kept`, plus a per-edge sweep and a concave-notch case.)

**4. Filtering happens in the sensor frame.**
Masking is applied to raw polar returns before any transform, so a mask cannot inherit an error in a mount that has never been measured. (`the_mask_is_applied_before_the_transform`.)

**5. Excessive masked-ray coverage makes the frame unhealthy.**
This is the least obvious guard and the one worth the most review attention.

`scan_confidence` had to stop counting masked rays in its denominator — a ray the robot's own body blocks cannot report the world, and counting it as a failed return means arming a *correct* mask costs confidence. A safety feature whose only visible effect is to push the corridor unhealthy never gets enabled, and a feature nobody enables protects nobody.

But that exclusion opens a hole the metre-based radius bound does not close: a thin annulus hugging the sensor stays comfortably inside the vertex-radius limit while subtending every ray in the scan, shrinking the denominator until a handful of survivors read as full confidence. `MAX_SELF_MASKED_RAY_FRACTION` (0.5) fails the frame closed before that can happen. (`a_mask_cannot_buy_confidence_by_shrinking_the_denominator` constructs exactly that annulus.)

**6. Production enablement requires verified mount provenance.**
`MountProvenance` records how the sensor-to-base transform was established. `assert_mount_verified()` refuses anything but `Measured`, so a production configuration cannot inherit a guess. (`an_unverified_mount_is_refused_for_production`.)

**7. No mask becomes active without the hardware capture procedure.**
There is no default mask and no way to get one implicitly — `SelfFilterMask` has no `Default`, and construction is fallible and explicit. The R2's real geometry does not exist in this branch because the capture has not been run.

## The standing transform bug — not introduced here, exposed here

Worth separating from the rest of this PR, because it is older than this branch and wider in scope than self-filtering.

**Taj treats the sensor origin as the base origin.** There is an implicit identity transform between the LiDAR and `base_link` that nothing has ever verified. If the R2's LiDAR is mounted forward of `base_link` — and it almost certainly is — then **every object Taj reports is displaced by that offset**, and no test in this repository would notice.

This affects all object locations, corridor geometry and clear-distance calculations. It is not confined to the self-filter path; it was simply invisible until the calibration work forced the question of where the sensor actually is.

This PR does not fix it. It makes the assumption explicit (`MountProvenance::AssumedIdentity` is the current, honest value), gates production on a real measurement, and specifies the tape-measure procedure that produces one — including the instruction to measure the built robot rather than read the URDF, and to record any discrepancy as a finding in its own right.

`a_measured_mount_offset_moves_the_reported_geometry` pins the consequence: supplying a measured offset shifts reported positions by exactly that offset. That is intended, not a regression — if the sensor is 0.12 m forward, the old positions were wrong by 0.12 m. The test exists so the day someone measures it, the change surfaces in CI rather than on a robot.

The number that measurement produces may require a follow-up to the corridor and clear-distance geometry. Reviewers should not assume it will be zero because that is what the code assumes today.

## Also in this change

`ingest_scan` is the single funnel every downstream stage consumes, and it was dropping rays silently. It now returns a per-reason `RejectionTally` — `not_a_number`, `below_min_range`, `beyond_max_range`, `self_masked`.

NaN is counted apart from `+inf` deliberately: `+inf` is the ROS no-return convention and means open space, so filing it under a fault counter would make an empty room read as a failing LiDAR. That distinction matters for #1211, which has to tell a broken sensor from a clear one.

The tally rides `TajPerception` to the sidecar wire as observability and is deliberately **outside** the evidence digest — it changes no verdict, and the one way self-filtering can affect safety already reaches the digest through confidence. (`the_tally_stays_out_of_the_evidence_digest` proves this rather than asserting it in a comment.)

## Tests

45 new tests: 31 unit (geometry, bounds, provenance, digest), 11 pipeline, 3 sidecar.

One is worth calling out. `the_survival_test_is_not_vacuous` proves that the same pipeline and the same scan shape *does* remove the object once it moves inside the mask — so "a real obstacle survives filtering" is meaningful rather than trivially true because filtering never fired. Without its non-vacuity pair, the survival assertion would pass just as happily against a no-op.

Two of my own tests were wrong before they were right, and both corrections are in the branch rather than papered over. One asserted masked rays never depress confidence — arithmetic does not support that, since masked rays were valid returns and removing them from both sides lowers the ratio slightly; it now asserts the bound that actually holds. Another assumed a NaN ray would reach the sidecar tally, but `validate_perception_request` refuses the whole frame first — stronger behaviour than I had assumed, so the test now pins that instead.

`cargo test --workspace` green. `cargo clippy` clean on both changed crates (the one warning in `kirra-sidecars` is pre-existing in `capabilities.rs`, untouched here). Quality guardrails, safety-constants and re-export-shim ratchets all green.

## What is deliberately not here

The physical capture. Which returns *are* the robot is a measurement, not an estimate, and it needs the hardware on blocks with a tape measure. The procedure is written up in `docs/hardware/R2_SELF_FILTER_CALIBRATION.md` — capture across centre and full lock, identify every cluster by hand and photograph it, refuse to mask any cluster that cannot be physically named, then verify by stepping a 100 mm box outward from the mask boundary in 10 mm increments and confirming an object appears at every position outside it.

Shipping disarmed is not free. The R2 currently reports its own mast as an obstacle, so the corridor is narrower and the speed cap lower than the world justifies. That is availability lost to caution — and it is the right trade to hold until the capture exists, because the other error is the robot not seeing something real.

---


justinlooney:
## Rebased onto main (now containing #1222 + #1223 + #1224)

Force-pushed `6802909e`. Three conflicts in `crates/kirra-sidecars/src/taj.rs`, all resolved as a **union** — the two changes are orthogonal and both belong:

| Site | This PR adds | #1223 added | Resolution |
|---|---|---|---|
| `invalid_perception_response` | `rejected: RejectionOut::default()` | four `sensor_*` fields + `apply_sensor_health` | both fields, then the liveness stamp |
| `PerceptionResponse` struct | `rejected` + the `RejectionOut` wire type + `From` impl | `sensor_liveness` / `sensor_fault` / `sensor_measured_hz` / `publisher_count_observed` | both field sets |
| response construction | `rejected: perception.rejected.into()` | four `sensor_*` initialisers | both |

The two features touch the same struct but nothing else: the rejection tally is per-ray accounting from `ingest_scan`, and the liveness verdict is a statement about the stream. Notably both landed on the same side of the same line — **each is deliberately outside the evidence digest**, for the same reason: neither changes what the sensor showed.

Verified after the rebase, not assumed:

- Self-filter intact — 31 `self_filter` module tests, 37 self-filter-related, 3 sidecar
- 128 sidecar tests (125 from #1223 + 3 here), 140 taj tests
- `cargo test --workspace` green
- clippy clean on both changed crates
- quality-guardrails, safety-constants, re-export-shim and mick-actuation-fence ratchets green

Nothing in the safety argument changed. The asymmetry this PR is built around — self-filtering is the one Taj operation that *removes* evidence — is untouched by the liveness work, and the default is still disarmed pending the hardware capture.

---

PR #1223, opened 2026-07-28
feat(taj): wire R2 scan liveness into perception cap

justinlooney:
> #1222 proved the watchdog logic; nothing called it. A frozen LiDAR still drove the robot. This connects it.

**Stacked on #1222** — based on `claude/r2-sensor-watchdogs-1211`, so the diff here is the wiring commit alone. Merge #1222 first; this retargets to `main` automatically. Independent of #1221.

## The path

```
raw /scan observation
        ↓
liveness watchdog tick
        ↓
sensor-health reason / cap
        ↓
existing Taj perception result
        ↓
perception_cap composition (min — tighter always wins)
```

**Publisher count is observed where it can be.** The ROS subscriber holds the graph handle, so `perception_governor` reads `get_publisher_count()` and carries it as data. The HTTP sidecar cannot inspect a ROS graph and does not pretend to — no graph inspection was pushed into the Rust state machine. An absent count is reported as `publisher_count_observed: false` rather than read as zero, so the check is skipped on a fact the caller could not know instead of firing on one. Timestamp, fingerprint, valid-ray ratio and rate stay in the pure watchdog.

**Fingerprinting is imported, never reimplemented.** The freeze check's entire discrimination between a frozen buffer and a static scene rests on hashing raw IEEE bits; a second implementation in the sidecar would be free to quantise and silently destroy it.

## Ordering that carries weight

The tick runs **before** the duplicate-stamp short-circuit. A driver replaying one buffer produces exactly the duplicate case, so a tick placed after it would never see the freeze it exists to catch. The cached response served on that path gets this tick's verdict re-applied — otherwise a frozen driver is handed the healthy cap computed before it froze, which is the bug.

## Three problems that only appeared against the real system

These are the parts worth reviewing closely. None were visible from the pure module.

**1. Two consumers post the same scan.** `perception_governor` and `occy_doer` both subscribe to `/scan` and both POST what they receive, so every frame reaches this service twice — which is why the duplicate short-circuit already existed. Counting the second copy as a fresh observation trips the frozen-stamp check on a perfectly healthy robot. **This would have grounded the R2.**

A re-post is now observed as *an arrival carrying no new frame*. A frozen driver still fails closed, because it also produces no new frames: the silence budget expires and it surfaces as `SENSOR_NO_MESSAGES` — "nothing new arrived", which is precisely true of a driver replaying one buffer. Detection preserved, false positive removed. The trade is that `SENSOR_FROZEN_TIMESTAMP` is no longer reachable *from this call site*; the condition is still caught, under a different and equally accurate code.

**2. Diagnostics caused the fault they diagnosed.** `kirra_doctor`'s profile-digest probe, `inspect_corridor` and `rabbit_ask` all POST canned scans with a fixed `stamp_ms` on every run, and share service state with the live stream. Unflagged, running the doctor would trip `SENSOR_FROZEN_FRAME` and leave the real robot owing a recovery streak afterwards. They now set `diagnostic_probe`, which excludes them from liveness accounting.

The flag can only *decline to observe* — it never sets or clears a fault, and never relaxes a bound. It also does not mask silence: ten probes with no live frames between them is a genuine one-second outage and still faults.

**3. A fault left its own outage in the rate window.** The gap created *by* a fault kept the measured rate under the floor long after the sensor recovered, so recovery took "until the outage ages out of the window" rather than the configured streak — the sensor punished twice for one fault. A stream break now clears the window. A rate fault **keeps** its window, because that fault *is* the window's verdict and erasing the evidence would make a genuinely starved sensor oscillate between reporting the rate fault and reporting nothing measurable.

## Two deliberate scope choices

**Startup begins unavailable.** `WarmingUp` now caps (it did not in #1222). A sensor observed once has not evidenced that it publishes at a rate, and "publishes at all" is not "publishes at rate".

**The stateless `handle_perception` keeps the watchdog disarmed.** This is correctness, not a compatibility concession: liveness is a property of a *stream* — frozen stamps, repeated buffers and measured rate all mean "compared to the frames before this one". A one-shot call has no frames before it, so an armed watchdog there would assert a verdict from one sample with no continuity to derive it from, and would cap every caller forever. `sensor_liveness: "disarmed"` is the truthful answer for a single-shot evaluation.

## Digest and cache ordering

The evidence digest is computed **before** the verdict is stamped, and the cache stores the pre-verdict response.

The digest identifies the perception *evidence* — what the sensor showed and what Taj made of it. Liveness is a verdict about whether that evidence is *current*, which is orthogonal: the same scan bytes produce the same evidence whether they arrived first or third in a frozen run. Folding it in would make a frame's identity depend on the history of unrelated frames, changing what a downstream consumer binds commands to. Same line #1210 drew for the rejection tally. `kirra_doctor` reads `profile_digest` (request-derived), unaffected either way.

Caching pre-verdict matters for the duplicate path: a stale `Some(0.0)` baked into the cache would compose with `min` forever, so a transient fault would outlive itself.

## Armed by default here

Unlike the library default. `kirra_trajectory::sensor_watchdog` defaults off because it is a library with unknown consumers; this sidecar **is** the R2 deployment #1211 was filed about, and shipping the wiring disarmed would leave the failure exactly where the issue found it. The service logs which posture it started in.

`KIRRA_SENSOR_WATCHDOG_ENABLED=0` is the opt-out and the immediate mitigation if bring-up shows false trips before thresholds are characterised on hardware.

## Tests

13 end-to-end, all with injected scans and an injected clock — **no sleeps**, since wall-clock waits would make a timing gate's own tests the flakiest thing in CI. `handle_perception_tracked_at` takes the clock; the binary reads it.

Frozen stream floors the cap (with a precondition assertion that the stream was serving motion first, so the test cannot pass vacuously) · recovery takes the exact streak, floored for every frame before it · a healthy camera cannot lift the floor · the cached duplicate carries the verdict · two consumers stay healthy · probes cannot trip the freeze but cannot mask silence either · malformed scans land on the existing reason codes · a caller's bad `margin_m` is not charged to the sensor · disarmed is unchanged · the verdict stays out of the evidence digest.

One test caught a real distinction while being written: ten probes with no live frames between them *is* a one-second outage, and the watchdog was right to fault it. The test was wrong, not the code.

`cargo test --workspace` green · 292 ROS package tests · 414 robot tests · clippy clean on both changed crates (the one `kirra-sidecars` warning is pre-existing in `capabilities.rs`) · quality-guardrails, safety-constants, re-export-shim and mick-actuation-fence ratchets green.

## After this, #1211 can close

Remaining work is tracked separately, as agreed:

- udev / device identity binding
- camera watchdog parity
- hardware threshold tuning (the 5 Hz / 0.15 defaults are deliberately loose placeholders)
- the #1210 self-filter mount measurement and mask capture

---

### DIFF
--- a/crates/kirra-sidecars/src/bin/taj_service.rs
+++ b/crates/kirra-sidecars/src/bin/taj_service.rs
@@ -16,7 +16,22 @@ use std::net::{TcpListener, TcpStream};
 use kirra_sidecars::capabilities::Capabilities;
 use kirra_sidecars::http::{read_request, respond, respond_error};
 use kirra_sidecars::net::{allow_nonlocal_from_env, enforce_bind_policy};
-use kirra_sidecars::taj::{handle_perception_tracked, PerceptionRequest, TrackedPerceptionState};
+use kirra_sidecars::taj::{
+    handle_perception_tracked_at, watchdog_armed_from_env, PerceptionRequest, SensorWatchdogConfig,
+    TrackedPerceptionState,
+};
+
+/// Wall clock in ms, for the watchdog's ARRIVAL observation.
+///
+/// Read here in the binary rather than inside the perception core so the core
+/// stays a pure function of its inputs and every test can inject a clock instead
+/// of sleeping.
+fn now_ms() -> u64 {
+    std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_millis() as u64)
+        .unwrap_or(0)
+}
 
 fn serve(mut stream: TcpStream, perception_state: &mut TrackedPerceptionState) {
     let req = match read_request(&mut stream) {
@@ -44,8 +59,12 @@ fn serve(mut stream: TcpStream, perception_state: &mut TrackedPerceptionState) {
             Ok(p) => respond(
                 &mut stream,
                 "200 OK",
-                &serde_json::to_string(&handle_perception_tracked(&p, perception_state))
-                    .unwrap_or_else(|_| "{}".into()),
+                &serde_json::to_string(&handle_perception_tracked_at(
+                    &p,
+                    perception_state,
+                    now_ms(),
+                ))
+                .unwrap_or_else(|_| "{}".into()),
             ),
             Err(e) => respond(
                 &mut stream,
@@ -74,7 +93,21 @@ fn main() {
     println!("Taj perception service on http://{addr}  (POST /perception, GET /health)");
     // Single-threaded service: one deterministic tracker state persists
     // across consecutive POST /perception requests.
-    let mut perception_state = TrackedPerceptionState::new();
+    let armed = watchdog_armed_from_env();
+    if armed {
+        eprintln!(
+            "taj_service: scan-liveness watchdog ARMED (#1211) — a frozen, starved \
+             or blinded /scan floors the speed cap. Set \
+             KIRRA_SENSOR_WATCHDOG_ENABLED=0 to opt out."
+        );
+    } else {
+        eprintln!(
+            "taj_service: scan-liveness watchdog DISARMED — a frozen /scan will \
+             NOT be detected. This is the pre-#1211 behaviour."
+        );
+    }
+    let mut perception_state =
+        TrackedPerceptionState::with_watchdog(SensorWatchdogConfig::r2_lidar(), armed);
 
     for stream in listener.incoming() {
         match stream {
--- a/crates/kirra-sidecars/src/bin/taj_service.rs
+++ b/crates/kirra-sidecars/src/bin/taj_service.rs
@@ -45,12 +45,23 @@ fn serve(mut stream: TcpStream, perception_state: &mut TrackedPerceptionState) {
         // healthy during R2 bring-up.
         ("GET", "/health") => {
             let caps = Capabilities::taj();
+            // #1211: report whether scan-liveness checking is ARMED, and whether it
+            // has latched terminal. A safety check that is switched off must not be
+            // indistinguishable from one that is on and passing — the same reason
+            // this endpoint already reports `contract` instead of a bare "ok".
             respond(
                 &mut stream,
                 "200 OK",
                 &format!(
-                    "{{\"status\":\"ok\",\"service\":\"taj\",\"contract\":{}}}",
-                    caps.contract
+                    "{{\"status\":\"ok\",\"service\":\"taj\",\"contract\":{},\
+                     \"scan_liveness_watchdog\":\"{}\",\"scan_liveness_terminal\":{}}}",
+                    caps.contract,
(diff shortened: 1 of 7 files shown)


justinlooney:
## Pre-merge review focus — results

Working through the seven points found **one real defect** and **one missing surface**. Pushed as `3da0118a`.

### The defect: duplicate re-posts could advance the recovery streak (point 2)

The subtle one, and it was real. A re-post is not silence — the budget has not expired — so it reached the watchdog as a *clean* tick, and any clean tick advanced the streak.

That meant **a frozen driver's own repetitions earned its recovery**: the stream would recover from a freeze on the strength of the very repetitions that constitute the freeze. Combined with the dedupe change, a driver replaying one buffer would have alternated between faulting on the silence budget and being paid back down by its own re-posts.

Only a tick carrying a **new frame** may advance the streak now. A tick without one holds the streak where it is, still capped. Mutation-checked: disabling the guard reds `duplicate_reposts_cannot_advance_the_recovery_streak` and nothing else.

### The missing surface: a disarmed watchdog was only in a log line (point 6)

`GET /health` now reports `scan_liveness_watchdog: "armed"|"disarmed"` and `scan_liveness_terminal`. A safety check that is switched off must not be indistinguishable from one that is on and passing — the same reason this endpoint already reports `contract` rather than a bare `"ok"`. Verified against the running binary in both postures:

```
{"status":"ok","service":"taj","contract":2,"scan_liveness_watchdog":"armed","scan_liveness_terminal":false}
{"status":"ok","service":"taj","contract":2,"scan_liveness_watchdog":"disarmed","scan_liveness_terminal":false}
```

### The other five

| # | Property | Status |
|---|---|---|
| 1 | A probe cannot advance the recovery streak | Already inert (probes never reach `observe`), but nothing pinned it. Now tested: ten probes cannot pay down a five-frame streak. |
| 3 | Continuous re-posts fault as no-new-messages | Holds — `SENSOR_NO_MESSAGES` once the silence budget expires. |
| 4 | A real frame after duplicates resumes correctly | Holds — resumes `recovering` at 1, neither `healthy` nor stuck faulted, and completes normally from there. |
| 5 | Two clients cannot double the measured rate | Holds — re-posts record no arrival, so a double-posted 10 Hz stream measures ~10 Hz, not 20. Worth having: a 20 Hz reading would mask a sensor that had genuinely halved its rate while two clients kept it looking nominal. |
| 7 | A restart begins unavailable | Holds — `warming_up`, capped, until a second observation makes a rate measurable. |

### End-to-end against the real binary

Not just the unit tests — the actual service, over HTTP:

```
live stream:    liveness=healthy  cap=4.099
frozen driver:  liveness=faulted  fault=SENSOR_NO_MESSAGES  cap=0.000  healthy=False
```

`cargo test --workspace` green · 125 sidecar tests · 292 ROS package tests · 414 robot tests · clippy clean on both changed crates · all four ratchets green.

---

Issue #1212, opened 2026-07-28
[R2 perception 3/8] Move speed-cap stabilization into the Rust checker — it exists in Python, outside the safety authority


justinlooney:
Child of #1209. Audit phase 4. **Verified status: exists in the wrong layer.**

## Verified current state

The plan calls this "not demonstrated". It is demonstrated — in the wrong process.

A raw-versus-stabilized cap state machine exists in `ros2_ws/.../perception_cap.py` and `perception_governor.py`. That is Python, in the doer-side ROS layer, outside the Rust checker where the safety authority lives.

**This is a worse position than absent, because it looks done.** A reader of the ROS tree sees cap stabilization and concludes the behaviour is enforced. It is not enforced — it is proposed, by a component the architecture explicitly does not trust for safety. Nothing stops a doer-side change from relaxing it.

Missing regardless of layer: enter/exit hysteresis, limiting-object persistence through a missed scan, monotonic-timestamp recovery.

## Acceptance criteria

- [ ] `raw_speed_cap_mps` and `stabilized_speed_cap_mps` are distinct, both carried on the checker's own state, both logged.
- [ ] **Restrictive change is immediate.** A tightening cap applies on the frame it is observed. No smoothing on the way down, ever.
- [ ] **Release requires a clear streak** — a configured number of consecutive frames with no limiting object before the cap rises.
- [ ] **Release rate is bounded** — the cap climbs at a limited rate, not in one step, even after a satisfied streak.
- [ ] **Limiting-object persistence** — a confirmed hazard missing from exactly one scan is retained, not treated as departed. Ties to the coasting lifecycle already in `kirra-taj`.
- [ ] **Enter/exit hysteresis** on the geometric trigger, so an object oscillating across a boundary does not oscillate the cap.
- [ ] Recovery keyed on **monotonic** time, not wall clock. A clock step must not release the cap.
- [ ] All transitions logged with the limiting object's identity and the reason for the change.
- [ ] The Python path is removed or demoted to advisory with a comment stating the checker is authoritative — not left in place to drift.
- [ ] Gate test: alternating hazard/clear frames at the sensor rate cannot produce full-speed oscillation. Assert the achieved cap envelope, not just that a function was called.
- [ ] The Nominal-path WCET is unchanged — this must not enter the per-pose `validate_vehicle_command` hot path.

## Done when

[...]

PR #1225, opened 2026-07-28
feat(trajectory): move speed-cap stabilization into the checker

justinlooney:
> The state machine already existed — in Python, in the doer layer, outside the safety authority. That is worse than absent, because it *looks* done.

Checker half of #1212. **The wiring that makes it authoritative is not here** — see *What this does not yet do*. #1212 stays open.

## A change of authority, not of policy

`ros2_ws/.../perception_cap.py` has a working raw-versus-stabilized cap state machine. A reader of the ROS tree sees it and concludes the behaviour is enforced. It isn't — it's *proposed*, by a component the architecture explicitly does not trust for safety, and nothing stops a doer-side edit from relaxing it.

The port is faithful. The asymmetry, the jitter epsilon, the confirmation streak, the rate limit and every fail-closed arm are the Python's, checked against its 12 governor tests. What changes is who owns the decision.

## Down is immediate, up is earned

A tightening cap applies on the frame it is observed, with no smoothing, ever. Smoothing a restriction means driving at a speed the world no longer supports for as long as the filter takes to catch up — the filter would be inventing clearance perception just withdrew.

A loosening cap is earned twice: a streak of consecutive clear frames before the cap may rise *at all*, then a bounded climb rather than a step. One optimistic frame is evidence of one frame.

Same shape as the Degraded decel-to-stop doctrine (SS-002) and the sensor watchdog's recovery streak: instantly more conservative, gradually less.

## Three things the Python never had

**Limiting-object persistence.** A confirmed hazard absent from exactly one scan is retained rather than treated as departed. One missed detection is an ordinary dropout — `kirra-taj`'s coasting lifecycle already models the same thing — while two consecutive misses is evidence it actually left. Retaining forever deadlocks the cap at its floor; retaining zero frames lets a flickering detection release it. One frame is the smallest grace that separates a dropout from a departure.

**Enter/exit hysteresis on the release side.** The Python's epsilon is jitter tolerance on the way *down*; it does nothing about an object hovering exactly at the geometric boundary. A frame now counts as clear only if it exceeds the standing cap by a margin, so the release threshold sits strictly above the engage threshold.

**The limiting object's identity travels with the decision**, so a transition logs as "held at 0.8 by object 41" rather than an unattributable number.

## Two bugs of my own, both found by writing the tests

Worth reading, because both are the kind that pass review by inspection.

**The margin initially gated the climb as well as the evidence.** As the cap closed in on the raw value, frames stopped counting as clear and the rise stalled up to `release_margin` short — `stabilized == raw` was unreachable, and a robot in an open corridor would have sat permanently below the speed perception allowed. Silent, invisible availability loss. The margin now gates evidence only; an earned release finishes its climb.

**Fixing that the obvious way broke the other side.** Carrying the streak past completion meant one historic release licensed every later one, so a boundary-straddling object could walk the cap up indefinitely without ever clearing the margin again. The evidence is now *spent* when the climb completes. Both directions have regression tests.

## Tests — 31

The gate the issue asks for, stated as an envelope rather than a call check: **100 hazard frames interleaved with 100 clear frames at 10 Hz cannot walk the cap above the hazard bound.** Its non-vacuity partner proves the same stream *without* the hazard frames does rise, so the bound is constraining real behaviour.

Boundary cases use binary-exact values where the default can't express them: `1.0 - 0.03` comes back as `0.030000000000000027`, strictly greater than the threshold, so a test claiming to sit exactly on the epsilon has to pick values that actually can. Noted in the test rather than papered over.

**Mutation gate: 87 mutants, 84 caught, 3 unviable, 0 missed** (run locally — `kirra-trajectory` is in the gate's scope). `cargo test --workspace` green, clippy clean, all ratchets green.

## WCET

Perception-path only. Nothing in `validation.rs` calls it, and the Nominal per-pose path is byte-identical — this must never enter `validate_vehicle_command`, and it doesn't. O(1) per scan, at scan rate.

## What this does not yet do

The checker now *can* own the decision; it doesn't *yet*. Still to come, and the reason #1212 stays open:

- The Taj sidecar computing the stabilized cap and serving it, so `speed_cap_mps` carries the checker's number with `raw_speed_cap_mps` alongside for observability
- `perception_governor.py` relaying that instead of running its own state machine, with the Python demoted so it cannot drift

Same split as #1222 → #1223: logic first, then the wiring that makes it real. I'm not claiming the issue closed on the strength of the module alone.

---

### DIFF
--- a/crates/kirra-trajectory/src/cap_stabilizer.rs
+++ b/crates/kirra-trajectory/src/cap_stabilizer.rs
@@ -394,6 +394,13 @@ impl CapStabilizer {
         self.stabilized_mps = 0.0;
         self.clear_streak = 0;
         self.last_now_ms = None;
+        // Defensive, and deliberately unobservable: clearing `last_now_ms` above
+        // sends the next observation down the first-observation path, which
+        // overwrites `last_raw_mps` before anything reads it. No test can
+        // distinguish this line's presence today. It stays because it costs nothing
+        // and the first-observation path is not guaranteed to keep that property —
+        // a reader adding an early `raw_tightened` check would silently depend on
+        // it. Noted so nobody hunts for the missing test.
         self.last_raw_mps = None;
         self.limiting = None;
     }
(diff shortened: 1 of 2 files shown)


justinlooney:
## Review invariants → tests

Mapping the ten requested invariants found **two that were not provable**. Both were inside this module rather than the wiring, so they're fixed here.

| # | Invariant | Test |
|---|---|---|
| 1 | A hazard frame can immediately tighten the cap | `a_restriction_applies_immediately_and_is_never_smoothed` |
| 2 | Clear evidence is consumed when one climb completes | `a_completed_release_spends_its_evidence` |
| 3 | Old clear evidence cannot authorize future climbs | `a_completed_release_spends_its_evidence` · `falling_back_to_exactly_the_standing_cap_discards_clear_evidence` |
| 4 | One missing observation preserves identity for the single-scan grace only | `a_hazard_missing_from_one_scan_is_retained` |
| 5 | A second miss removes that persistence | `a_hazard_missing_from_two_scans_has_departed` |
| 6 | Hysteresis cannot let alternating frames ratchet above the hazard envelope | `alternating_hazard_and_clear_frames_cannot_oscillate_the_cap` + `the_oscillation_gate_is_not_vacuous` |
| 7 | **Reset/restart begins from the fail-closed state** | **was unprovable — see below** |
| 8 | NaN / negative time / non-monotonic time / malformed caps fail closed | `a_malformed_raw_cap_fails_closed` · `a_backwards_clock_fails_closed` · `an_oversized_frame_gap_fails_closed` · `negative_time_is_structurally_impossible` |
| 9 | **Identity absent when nothing currently or persistently limits** | **thin — see below** |
| 10 | The gate test bounds output, not that a function was called | `alternating_…` asserts the peak enforced cap over 200 frames |

### (7) There was no `reset()`

The Python has one; I hadn't ported it, so the invariant was untestable and the API was incomplete. Added, mirroring the Python — back to the initial cap, no evidence, no continuity, no limiting object.

The reasoning: a restart must not inherit a cap earned against observations this instance can no longer see. The evidence and the clock history are gone, so the permission they bought goes with them. Three tests: a fresh stabilizer begins at the floor, a reset discards an earned cap *and* its continuity (the next frame is `InitialHold` again), and a reset clears a retained limiting object.

**One hazard worth your call.** `initial_cap_mps` is configurable and non-negative, inherited from the Python. Any non-zero value means a fresh or reset stabilizer permits motion having observed **nothing** — which contradicts every other decision in this module, and strictly speaking breaks invariant 7 as you stated it. The default is 0.0, so the shipped behaviour is fail-closed. I documented it as a deliberate-exception knob rather than removing it, because dropping a field from ported behaviour is your call, not mine. Say the word if you'd rather it be hard-wired to zero.

### (9) Identity-absence had one assertion, now three

An unlimited stream never reports a limiting object · a reset clears a retained one · **a fail-closed arm clears it too**.

That last one is the interesting case: a stabilizer that has stopped trusting its input must not keep attributing the refusal to a specific object. A consumer logging "held by object 41" needs to be able to trust that 41 is real, and after a NaN cap it isn't — the attribution would be inherited from a frame the stabilizer just rejected.

### (8) Negative time

No runtime arm needed: `now_ms` is `u64`, so the type forbids what the Python validated at runtime. There's a test documenting that the guarantee is structural rather than checked, so nobody later "adds the missing validation".

---

37 tests · mutation gate **88 mutants, 85 caught, 3 unviable, 0 missed** · clippy clean · workspace green.

**Unrelated observation while verifying:** one full-workspace run showed 2 failures in `kirra-verifier --bin kirra_verifier_service`. That crate does not depend on `kirra-trajectory`, it passes 141/141 in isolation, and a second full-workspace run was clean — so they look intermittent under parallel load. I could not capture their names because the re-run passed, so I'm reporting this as an observation rather than a diagnosis. Pre-existing and outside this PR.

---

justinlooney:
## Open decision for reviewers: keep or drop the unobservable `last_raw_mps` reset

Flagging this as a **decision point, not a settled choice**, so it gets weighed rather than skimmed.

`reset()` clears five fields. Mutating each one individually and re-running the suite:

| Field removed from `reset()` | Tests that fail |
|---|---|
| `stabilized_mps` | 3 |
| `clear_streak` | 2 |
| `last_now_ms` | 3 |
| `limiting` | 3 |
| **`last_raw_mps`** | **0** |

That last row is not a missing test. Clearing `last_now_ms` sends the next observation down the first-observation path, which writes `last_raw_mps` before anything reads it — `raw_tightened` is evaluated strictly after that branch. The assignment is genuinely unobservable under the current state machine, so any test "covering" it would be asserting an internal detail rather than a behaviour, and would break the moment someone legitimately reordered the branch.

**The two defensible positions:**

- **Keep it** (current) — a `reset()` that visibly clears all historical state is easier to audit than one where a reader has to reconstruct why one field is safe to skip. The redundancy costs one line and one comment. It also stops being redundant the moment anyone adds an early `raw_tightened` check, at which point its absence would be a silent bug.
- **Drop it** — the new-epoch path guarantees overwrite-before-read, so the line is dead by construction, and dead code in a safety module has its own cost.

Current position is keep, with a comment in the source stating exactly why no test can distinguish it, so nobody hunts for the missing coverage or deletes it as unused. Happy to drop it if reviewers prefer.

Worth noting `cargo-mutants` does not generate statement-deletion mutants, so this is not a CI gate risk either way — it only showed up because I mutated the field-clears by hand.

## CI

In progress on the latest push (`Mutation gate` and `Branch Coverage` still running, everything completed so far green). Local: 44 tests, mutation gate 85 mutants / 82 caught / 3 unviable / **0 missed**, clippy clean, `kirra-trajectory` 254 tests green.

---

Issue #1226, opened 2026-07-28
[flaky] 2 intermittent failures in kirra_verifier_service under full-workspace parallel test runs


justinlooney:
**Status: observed, not diagnosed.** Filing now so it is not lost; I could not capture the test names because the reproduction attempt passed.

## What was seen

During routine verification on the #1225 branch, one `cargo test --workspace` run reported:

```
failures:
failures:
test result: FAILED. 139 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
error: test failed, to rerun pass `-p kirra-verifier --bin kirra_verifier_service`
```

Two tests in the `kirra_verifier_service` binary target failed.

## What was checked

- **Passes in isolation.** `cargo test -p kirra-verifier --bin kirra_verifier_service` → **141 passed, 0 failed**. Note 139 + 2 = 141, consistent with the same two tests passing when run alone.
- **Did not reproduce.** A second full `cargo test --workspace` run on the identical tree was clean — no failures block at all.
- **Not caused by the branch under test.** #1225 touches only `crates/kirra-trajectory`, which the verifier binary does not depend on. (`kirra-trajectory` appears in the root `Cargo.toml` only as a workspace *member*, not as a dependency edge.)

## Why the names are missing

The capture filter on the first run matched the `failures:` header lines but not the test names that follow them, and by the time a targeted re-run was attempted the failure had gone. Re-running with `--no-capture` or `-- --test-threads=N` under concurrent workspace load is the obvious next step.

## Hypothesis, explicitly unconfirmed

The only distinguishing condition observed is **parallel load**: the failure appeared when many test binaries were executing concurrently and not when the binary ran alone. That points at a shared-resource or timing sensitivity — this binary's suite includes TLS/mTLS live-handshake tests that bind sockets, systemd-notify tests, and store-offload tests, all of which are plausible candidates. **This is a guess and should not be treated as a finding.**

## Why it matters beyond the noise

A flaky test in the safety service is worse than a flaky test elsewhere, because the standard response to intermittent red is to re-run until green — which is indistinguishable from re-running until a real intermittent fault hides. Whatever the cause, it should end up either fixed or explicitly quarantined with a reason, not tolerated.

## Suggested next step

Run the full workspace suite in a loop (10–20 iterations) capturing full output, and record the names the moment it reproduces. Until then there is nothing actionable beyond this record.

## Not blocking

#1225 is unrelated and should not wait on this.

[...]

PR #1227, opened 2026-07-29
feat(taj): migrate cap-stabilization authority to the checker

justinlooney:
> #1225 gave the checker a stabilizer. Nothing called it — the Python state machine was still the live authority. This makes the checker's number the one that reaches the wheels.

**Completes #1212.** Off `main` (which has #1225).

## The migration, end to end

```
Taj raw cap
  → Rust stabilizer owns state          (#1225)
  → sidecar serves raw + stabilized + limiting identity
  → perception_governor relays
  → Python stabilizer DELETED
```

## What the wire carries now

`speed_cap_mps` becomes the **stabilized** value — the enforced number — with `raw_speed_cap_mps` alongside, plus `cap_reason`, `cap_clear_streak` and `cap_limiting_object`.

Three numbers matter to an operator: what perception saw, what liveness allowed, what is enforced. So `raw_speed_cap_mps` stays perception's own geometry and is **not** overwritten by the liveness floor — an early draft did overwrite it, and a warm-up frame then reported `raw = 0`, which is exactly the collapse that made the Python's behaviour invisible.

## Ordering, and why it composes

Stabilization runs **last**, over the liveness-adjusted cap. Feeding a sensor fault's zero through the stabilizer makes a fault behave exactly like a hazard — restrict instantly, re-earn the full streak on recovery. That is what the Python did with `effective_raw_cap = 0.0 if not healthy`.

It also means **a recovering sensor must earn both streaks**. "The sensor is producing frames again" and "the envelope may widen" are different claims, and `a_recovering_sensor_must_earn_both_streaks` pins that the cap is still floored at the moment liveness reports healthy. Three existing liveness tests were scoped back to their actual subject rather than weakened — they were asserting cap release as a side effect of liveness recovery, which is no longer true and shouldn't be.

## Limiting-object identity

Reported only when an **object** bound the clear distance, not when the corridor geometry did. Naming one otherwise would have the persistence logic hold a cap against a hazard that never existed.

The nearest-object fold previously discarded which object won; it now tracks the id, with **ties broken by lower id** so two objects at equal distance cannot alternate the reported identity frame to frame — which would look exactly like a dropout to the persistence logic and hold the cap for a grace window on nothing.

## Re-posts and probes

Serve the held cap without advancing the stabilizer, same rule as the watchdog. Otherwise two consumers posting the same scan inflate the frame rate the rise limiter divides by, and a two-consumer deployment climbs **twice as fast** as a one-consumer deployment on identical perception. `a_repost_does_not_advance_the_stabilizer` runs single- and double-posted streams side by side and requires identical caps.

## Two gates, not one

Stabilization is gated separately from the liveness watchdog (`with_options`). They answer different questions, so disabling one must not silently disable the other — the disarmed-watchdog test now turns both off explicitly rather than asserting the absence of two features while naming one.

The stateless `handle_perception` disables both. Stabilization is a property of a *stream*: a restriction is "tighter than the last frame", a release is "clear for N frames". A one-shot call has no previous frame, so an armed stabilizer would hold every caller at the floor forever while asserting a temporal judgement it has no history to make. It reports `cap_reason: None` rather than a stabilized number it did not compute.

## Python: relay, then delete

`perception_governor` relays and computes nothing. The relay decision is extracted into `perception_cap.resolve_stabilized_cap` so it is unit-testable without ROS, and it **fails closed** when the field is absent or malformed.

That last part matters more than it looks: an older Taj that does not serve a stabilized cap is not a licence to fall back on the raw one. Falling back would silently restore pre-#1212 behaviour on exactly the deployments that hadn't been upgraded — the worst place for it and the hardest to notice.

`SpeedCapGovernor`, its config, its decision type and its **15 tests are deleted**, not demoted to advisory. An unused copy of a safety state machine drifts silently: its tests keep passing while nothing enforces it, which is the exact condition that made the original placement a problem.

## Tests

134 sidecar (7 new) · 281 ROS (4 new for the relay, 15 governor tests removed) · 414 robot · workspace green · clippy clean · guardrails and mick-actuation-fence green.

**Three of the new tests started out asserting the wrong thing**, and the corrections are worth reading:

- A restriction to a hazard still *above* the standing cap registers as `CAP_RESTRICTED` without moving the served number. That's correct — the enforced cap was already more conservative than the new hazard requires — but it proves nothing about immediacy, so the test now uses a hazard below the standing cap.
- A post-reset rise of one bounded step is the normal climb continuing, not a reset handing back a cap. The assertion now bounds the *jump* rather than forbidding movement.
- `raw_speed_cap_mps` read zero on a warm-up frame, which surfaced the liveness-overwrite issue described above.

## Note on an unrelated flake

The full-workspace run showed the same 2 intermittent `kirra_verifier_service` failures recorded in #1226 — second reproduction, identical `139 passed; 2 failed` signature, and it passed clean on re-run again so the names still aren't captured. Unrelated to this diff and not blocking.

---

### DIFF
--- a/crates/kirra-sidecars/src/taj.rs
+++ b/crates/kirra-sidecars/src/taj.rs
@@ -795,8 +795,14 @@ pub struct PerceptionResponse {
     /// it.
     pub stabilized_speed_cap_mps: f64,
     /// Why the stabilizer produced this value (`CAP_RESTRICTED`, `CAP_RISING`,
-    /// `CAP_TIME_GAP`, …), or `None` when stabilization did not run for this
-    /// request (stateless call, diagnostic probe, or a rejected frame).
+    /// `CAP_TIME_GAP`, `CAP_REPLAY_HELD`, …).
+    ///
+    /// `None` means **this sidecar performed no stabilization at all** — a stateless
+    /// call or a build that does not stabilize. It is a statement about the
+    /// authority that produced `speed_cap_mps`, not about the outcome, which is why
+    /// a consumer may treat it as unusable and fail closed. Every path where the
+    /// stabilizer DID author the number carries a code, including a replay of a cap
+    /// it is holding from an earlier frame.
     pub cap_reason: Option<&'static str>,
     /// Consecutive clear frames accumulated toward a release.
     pub cap_clear_streak: u32,
@@ -999,6 +1005,48 @@ fn is_forward_object_candidate(
         && y_m.abs() <= lane_half_m
         && y_m.atan2(x_m).abs() <= MAX_FORWARD_OBJECT_BEARING_RAD
 }
+
+/// Pick the nearest `(id, distance)` candidate, or `None` when none is usable.
+///
+/// Extracted from the perception path so the ordering is testable on its own. It has
+/// two properties the caller depends on and neither is obvious from the call site:
+///
+/// **Non-finite distances are dropped BEFORE the comparison**, not after a winner is
+/// chosen. The ordering below is a total order over finite values only; a NaN reaching
+/// it compares `false` against every candidate and therefore WINS every match arm.
+/// Filtering afterwards would discard that winner and report no nearest object at all
+/// — so one malformed distance would hide every real obstacle in the frame. Today the
+/// caller's inputs are already proven finite (`is_forward_object_candidate` rejects a
+/// non-finite `x`, and the near-edge lookup is finite-filtered), which is exactly why
+/// the guard belongs here: the invariant lives with the code that needs it rather than
+/// depending on two callers upstream continuing to hold it.
+///
+/// **Ties break on the lower id, compared exactly.** Two objects at an identical
+/// distance must not alternate the reported identity frame to frame — to the
+/// stabilizer's limiting-object persistence that alternation is indistinguishable from
+/// a dropout. The comparison is never epsilon-based: an approximate "near enough"
+/// predicate is not transitive, so a three-candidate frame could order inconsistently
+/// depending on iteration order. Two distances differing in the last bit are two
+/// distances, and the nearer one wins.
+#[must_use]
+fn nearest_candidate(candidates: impl Iterator<Item = (u64, f64)>) -> Option<(u64, f64)> {
+    candidates
+        .filter(|(_, d)| d.is_finite())
+        .fold(None::<(u64, f64)>, |acc, (id, d)| match acc {
+            Some((bid, bd)) if bd < d || (bd == d && bid <= id) => Some((bid, bd)),
+            _ => Some((id, d)),
+        })
+}
+
+/// `cap_reason` for a cap served WITHOUT the stabilizer observing the frame — a
+/// duplicate re-post, served the value the stabilizer is currently holding.
+///
+/// Sidecar-local rather than a [`kirra_trajectory::cap_stabilizer::CapReason`]
+/// variant, because it is not a decision the stabilizer made. The stabilizer never
+/// saw this frame; the sidecar chose not to show it one. Putting it in the checker's
+/// enum would oblige every consumer of that enum to reason about a state the state
+/// machine cannot enter.
+const CAP_REPLAY_HELD: &str = "CAP_REPLAY_HELD";
 
 const MAX_TRACKING_GAP_MS: u64 = 500;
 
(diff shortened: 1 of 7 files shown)


justinlooney:
Worked through the seven gates. Five surfaced real defects; two were satisfied but untested, and are now tested. Pushed as `1a251e67`.

**What the gates found**

*Gate 6 — limiting identity must match the stabilized cap.* Two incoherences, both mine.

The replay path (`apply_cap_without_observing`) served the cap the stabilizer was holding while stamping `cap_reason: None` and leaving `cap_limiting_object` at `None`. That describes a held cap as unstabilized and bound against nothing — which reads as an open corridor, the opposite of the truth, to anything attributing the restriction. It now carries `CAP_REPLAY_HELD` and the retained identity, via a new `CapStabilizer::limiting_object()` for callers that serve a held cap without observing a frame.

Separately, the fresh path derived `nearest_object_id` and `nearest_object_m` from the same `Option` but then applied `.filter(|d| d.is_finite())` to only one of them. Had it ever fired it would have named an object bounding a cap that reported nothing bounding it. They are now one destructured value.

*Gate 5 — float edge cases in the tie-break.* The ordering `bd < d || (bd == d && bid <= id)` is total over finite values only. A NaN compares `false` in both directions, so it wins every match arm — and the post-filter would then discard the winner and report **no nearest object at all**. One malformed distance would have excluded every real obstacle in the frame from the clear-distance computation. The filter now runs *before* the fold.

This is unreachable today: `is_forward_object_candidate` already rejects a non-finite `x`, and the near-edge lookup is finite-filtered. That is exactly why the guard belongs with the ordering rather than depending on two callers upstream continuing to hold it. I extracted the fold to `nearest_candidate` so the property is testable directly — a synthetic scan cannot produce a NaN distance or two bit-adjacent ones on demand, which are the cases where an ordering stops being total.

On the epsilon question: no approximate comparison was introduced, and one would be wrong here. It is not transitive, so a three-candidate frame could order inconsistently depending on iteration order. There is a test asserting every rotation and reversal of a five-element set yields the same winner.

*Gate 3 — relay fail-closed for every unusable form.* You were right that this was the hole. `resolve_stabilized_cap` accepted `stabilized_speed_cap_mps` without checking authorship — and a sidecar that does **not** stabilize still populates that field, with the raw cap. The number alone cannot distinguish the two, so an un-upgraded deployment would have had pre-#1212 behaviour restored silently, with a value that looked entirely reasonable. `cap_reason` is now required before the value is read.

`UNSTABILIZED` and `MALFORMED` are reported separately because the operator action differs — only one is fixed by a restart. An unusable clamp bound is now a fault too: clamping against NaN passes the stabilized value through untouched.

*Gate 4 — authority absent, not merely unused.* The four cap parameters were still declared, and `cap_restriction_epsilon_mps` was still read into a field nothing consumed. rclpy silently ignores an override for an undeclared parameter, so the YAML entries were tuning that appeared to apply and did nothing — an operator tuning them would see the value in `ros2 param get` and conclude the policy changed. All four are gone from the node and the config. A guard test asserts no cap state machine returns to the Python module. A dead `STABILIZED_MISSING` alias I had left behind is also gone.

**What the gates confirmed**

*Gate 1 — coherent snapshot.* The ordering was already right: the response is cached **before** the liveness verdict and before stabilization, so a replay applies both fresh rather than serving a stale `Some(0.0)` baked into the cache. I corrected a comment that claimed three values reach the wire; two do. The liveness-adjusted intermediate is recoverable from `sensor_fault`/`sensor_liveness`, and carrying perception's own proposal is the part that matters.

*Gates 2 and 7 — composite latency, and the boundary as one sequence.* Both held, neither was tested. Now: after a liveness fault, motion returns only once **both** the watchdog streak and the stabilizer's confirmations plus bounded climb are satisfied — measured at **17 frames, ~1.7 s at 10 Hz**, against a 5-frame (~0.5 s) liveness streak. That composite is the figure an integrator needs and neither gate produces alone, so it is recorded in the test, in a bound, and in `AOU-CAP-STABILIZER-001` along with the deployment obligation the relay change creates.

Gate 7's test walks warm-up → climb → hazard → recovery → replay as one sequence, asserting at every step that the served number is the stabilized one and never exceeds perception's proposal. It carries its own non-vacuity check: if the stabilized cap had merely tracked the raw one throughout, every assertion would hold while nothing was being stabilized.

**Verification.** Each new assertion was mutation-checked by breaking what it protects — removing the pre-fold finite filter, reverting the replay reason to `None`, clearing the replay identity, and flipping the tie-break each fail a test. 147 sidecar tests, 285 ROS tests, 699 Python tests, workspace clippy clean.

**On #1226.** Not reproduced in 18 attempts (6 serial, 12 under 4× process contention) — `141 passed; 0 failed` every time, so the two names are still uncaptured. A full-workspace sweep is running now under the load profile where it originally appeared, configured to stop and preserve output on first failure rather than rerun to green. I have not treated the green runs as evidence the original failure was spurious.

---

Issue #1214, opened 2026-07-28
[R2 perception 5/8] Close the evidence-binding residue — verify monotonic expiry, add the Taj→Occy half, stable track generation IDs


justinlooney:
Child of #1209. Audit phase 12. **Verified status: substantially complete — the plan understated this.**

## Verified current state — read this before planning work

The source plan calls evidence binding "partial foundation; major work remains" and makes it PR 2 of five. **Five of its seven "not demonstrated" items are implemented and tested.**

| Plan item | Repository |
|---|---|
| `PerceptionFrameId` with scan/camera/tracker generations | `RosBoundCommandPayload` carries `scan_sequence`, `camera_sequence`, `tracker_generation` |
| evidence digest through Occy | `evidence_digest` in the signed payload |
| trajectory digest bound to perception digest | `proposal_digest` + `profile_digest`, one signed payload |
| rejection when a newer frame supersedes the plan | `RosBoundCommandRefusal::PerceptionFrameSuperseded`, with a test |
| single-use nonce at the motor consumer | `nonce` in the payload; consumer documented evidence-bound V2 only and refuses the older frame *because it names no evidence* |
| exact steering and velocity binding | in the signed payload |
| expiration on monotonic time | `expires_at_ms` — **source not verified** |

Treating this as major architecture would spend weeks re-deriving what already ships. The remaining work is narrow.

## Acceptance criteria

- [ ] **Verify `expires_at_ms` is monotonic, not wall-clock.** Trace the source at every producer. If it is wall-clock, a clock step back extends a stale proposal's life and a step forward kills a live one — either is a defect. Fix or document.
- [ ] Add a test that steps the clock backward and forward and asserts expiry behaviour is unaffected.
- [ ] **Close the Taj→Occy half of the frame contract.** The consumer end is bound; the perception-to-planner end is not. Taj's output must carry the same frame identity Occy echoes into the proposal, so a plan built on frame N cannot be presented as built on frame N+1.
- [ ] **Stable track generation IDs.** Tracks currently have persistent IDs but no generation counter that survives as a binding target. A proposal that names track 7 must be refusable when track 7 has been re-established as a different object.
- [ ] Confirm the nonce is genuinely single-use at the consumer — burned, not merely checked — with a replay test.
- [ ] Record what is *not* bound, explicitly, so the next audit does not re-derive it.

## Done when

[...]

PR #1228, opened 2026-07-29
fix(evidence-binding): close the #1214 residue — cold-start collision, clock steps, unbound proposals

justinlooney:
> #1214 said "partial foundation; major work remains." Five of its seven items already shipped. The work that remained was narrow — and two of the things it found were not on the list.

**Completes #1214.**

### The evidence binding never worked from a cold start

`PerceptionFrameId::invalid()` is all-zeros, and every consumer refuses a zero `tracker_generation` on exactly that basis — a default-constructed frame must not pass as perception evidence. The planner seam rejects it as `INVALID_PERCEPTION_FRAME_ID`.

Taj's **first live generation was also zero**. So a freshly-started sidecar served real frames carrying the sentinel value, and the planner refused them. The evidence-bound motion path could not work from a cold start at all; it only came alive if a tracker reset happened to advance the counter off zero, which needs a timestamp fault or a config change.

Confirmed by probing the running code, not by reading it:

```
PROBE first tracker_generation = 0
PROBE first scan_sequence      = 1
```

It failed closed — the robot held rather than moving on unbound evidence — which is exactly why it went unnoticed. **An inert safety binding and a working one look identical from outside as long as nothing is trying to move.**

The first live generation is now `1`, and `0` stays reserved. Both halves are pinned: a cold-started tracker must not report the sentinel, *and* the sentinel must remain distinguishable from every live frame. Without the second test the first could be "fixed" by making consumers accept zero, which would let a default-constructed frame authorize motion.

### A wall-clock step could silently extend a token's life

Traced both ends of the freshness comparison. The verifier signs `issued_at_ms`/`expires_at_ms` with `SystemTime::now()`; the motor consumer compares against `time.time()`, in a different process and usually on a different host.

Both being wall clock is **not a mistake** — the comparison needs a shared epoch, and a raw monotonic clock cannot provide one across processes. AOU-TIMESYNC-001 requires that source be synchronized *and* monotonic, and states the consequence when the second property fails:

> an out-of-domain or unsynchronized timestamp **silently disables the deadline barrier**: no fault is raised, the check simply stops meaning what it asserts

The gap is narrower than "wrong clock," and more interesting. A step **larger** than the 200 ms token lifetime is already fail-closed: a big backward step reads as `FutureIssued`, a big forward one as `Expired`. A step **smaller** than the lifetime tripped nothing — a 150 ms backward step just extended every token's life by 150 ms, so a command that should have expired stayed releasable. That is the "drift within the plausible range" the AoU says runtime checks do not catch.

`ClockStepGuard` cross-checks the wall clock against the monotonic one. Scheduling jitter, a slow frame and a long silence all move both clocks equally and produce no signal; a divergence means the clock *jumped rather than flowed*. On detection the frame is refused **before the core sees it**, so no sequence, nonce or perception watermark advances on a comparison that is no longer meaningful, and liveness starves into the SS-002 stop exactly as silence does.

**The hold is derived, not chosen.** After a step the clocks advance together again, so resuming immediately would release tokens minted under the old offset that are still inside their window — genuinely ambiguous, since the consumer cannot tell whether it or the signer moved. The hold runs for one maximum token lifetime on the *monotonic* clock, after which every pre-step token is expired under any offset. Exactly as long as the ambiguity lasts.

A regressed monotonic clock is reported as itself, not as a wall-clock step: it breaks the assumption the detector rests on, and mislabelling it would send an operator to the wrong system.

## One fail-open closed

The planner accepted a plan naming **no evidence** and answered 200 with a well-formed `proposal_digest` — the `compute_proposal_digest` `None` arm hashes a zero byte and produces a valid-looking digest for a plan that cannot say what world state it describes.

The ROS doer already declined to propose motion from such a plan, so nothing unsafe shipped. That is exactly why it was worth closing: **the refusal lived in the caller, not in the authority.** The planner is the one component that knows whether it had evidence, and a second caller written without that downstream check would have inherited a silent fail-open.

The field stays `Option` on the wire deliberately, so absence produces a coded `MISSING_PERCEPTION_FRAME_ID` rather than a generic deserialization error — a caller has to be able to tell "you sent no evidence" from "your JSON was malformed."

## Two acceptance criteria had wrong premises

Reported rather than worked around.

**Stable track generation IDs (AC4)** — already satisfied. `TajTracker::next_id` never recycles within a tracker instance, and *every* tracker replacement routes through `reset()`, which bumps the generation. Both triggers (timestamp fault, config change) hit that path, so `(tracker_generation, object_id)` is already a stable binding target. Pinned with a test rather than reimplemented.

**Nonce single-use (AC5)** — already a genuine burn, not a check: a strictly-advancing watermark giving single-use semantics in constant memory, with refusals leaving it untouched. What was missing was *tests*, so two were added:

- The **canonical replay** — the exact same bytes and token, presented twice. Every existing replay test mutates a field and re-signs, which assumes a capability the attacker does not have. It also distinguishes a burned nonce from a checked one: a validity check passes the second time, because the bytes are by construction as valid as they were the first. The test asserts the frame is still inside its window at refusal, so it cannot pass for the wrong reason if fixture timings drift.
- **A refused frame does not burn the identity it presented.** If refusals advanced the watermark, anyone able to inject one malformed frame could burn a range of sequence and nonce values and lock out the legitimate governor — turning replay protection into a denial-of-service primitive.

## What is NOT bound — `docs/safety/EVIDENCE_BINDING.md`

AC6, and the most useful artifact here. The chain reads stronger than it is: every hop validates **shape** and carries the value forward, and **no component recomputes a digest it received**.

Most importantly: the verifier signs whatever evidence identity the HTTP body supplies. It has no channel to Taj and cannot verify it. That is bounded by the actuator scope and the full envelope/posture/decel checks — the binding's job is to make evidence *attributable and non-replayable*, not to authenticate the perception stack — but the distinction needs to be written down rather than inferred.

Also recorded: the planner carries frame identity without using it; the enforced speed cap is deliberately outside the evidence digest (a verdict is not evidence); the bench minter binds nothing real; and the cold-start consequence that **restarting Taj alone is not a valid recovery action** — a long-lived consumer refuses the restarted sidecar's frames as superseded until it too restarts.

#1214's own text is the argument for writing this down: it estimated "major work remains" for an area where five of seven items already shipped. Re-derivation is expensive.

## Verification

74 release-token · 151 sidecar · 717 Python (18 new for the clock guard) · workspace green · clippy clean · fmt clean.

Every new assertion mutation-verified by breaking what it protects — inverting the tolerance check, measuring the hold on the wall clock, ignoring a monotonic regression, latching the hold, ignoring the verdict, running the guard after the core, reverting the generation fix, and flipping the tie-break each fail a test.

## Note on the branch

The remote branch held one already-merged commit (`9b4761be`, squashed into `main` as `e567d360` via #1221). `git cherry` confirmed it upstream and every file identical to `main` or superseded by later work, so the branch was restarted from `main` per the merged-PR procedure.

---

### DIFF
--- a/crates/kirra-release-token/src/ros_bound_command.rs
+++ b/crates/kirra-release-token/src/ros_bound_command.rs
@@ -603,6 +603,79 @@ mod tests {
         assert_eq!(gate.last_released_nonce(), Some(9001));
     }
 
+    /// #1214: the canonical replay — the EXACT same bytes and the EXACT same
+    /// token, presented again.
+    ///
+    /// This is the only replay an attacker who cannot forge the governor's
+    /// signature is able to mount: capture a frame that was legitimately
+    /// released and re-present it verbatim. Every other replay test in this
+    /// module mutates a field, which requires re-signing and therefore assumes a
+    /// capability the attacker does not have.
+    ///
+    /// It also distinguishes a BURNED nonce from a merely CHECKED one. A check
+    /// against a validity rule would pass a second time, because the bytes are
+    /// by construction as valid as they were the first time. Only consuming the
+    /// nonce refuses them — here via the strictly-advancing watermark, which
+    /// gives single-use semantics in constant memory rather than an unbounded
+    /// set of every nonce ever seen.
+    #[test]
+    fn a_verbatim_replay_of_a_released_frame_is_refused() {
+        let mut gate = gate();
+        let payload = payload();
+        let token = signed(&payload);
+        let bytes = payload.encode();
+
+        gate.release(&bytes, Some(&token), 10_050)
+            .expect("the first presentation is honest and releases");
+
+        // Same bytes, same signature, still inside the validity window: nothing
+        // about the frame itself has become invalid.
+        let replayed = gate.release(&bytes, Some(&token), 10_060);
+        assert_eq!(
+            replayed,
+            Err(RosBoundCommandRefusal::SequenceNotAdvanced {
+                presented: 10,
+                last_released: 10,
+            }),
+            "a captured frame must not be releasable twice"
+        );
+
+        // Still inside the window — so the refusal above is replay protection,
+        // not expiry doing the work. Without this the test would pass for the
+        // wrong reason as soon as the fixture's timings drifted.
+        assert!(10_060 < payload.expires_at_ms);
+
+        assert_eq!(gate.last_released_nonce(), Some(9001));
+        assert_eq!(gate.last_released_sequence(), Some(10));
+    }
+
+    /// A refused frame does not consume the identity it presented.
+    ///
+    /// The complement of the burn: if a refusal advanced the watermark, anyone
+    /// able to inject one malformed frame could burn a range of sequence and
+    /// nonce values and lock out the legitimate governor — turning replay
+    /// protection into a denial-of-service primitive.
+    #[test]
+    fn a_refused_frame_does_not_burn_the_identity_it_presented() {
+        let mut gate = gate();
+        let honest = payload();
+
+        // Refused for a bad signature — signed with the wrong key.
+        let impostor = SigningKey::from_bytes(&[7u8; 32]);
+        let forged = issue_ros_bound_command_release(&honest, &impostor);
+        assert!(gate
+            .release(&honest.encode(), Some(&forged), 10_050)
+            .is_err());
+
+        // The same sequence and nonce, honestly signed, must still work.
+        let token = signed(&honest);
+        assert_eq!(
+            gate.release(&honest.encode(), Some(&token), 10_050),
+            Ok(honest),
+            "a rejected impostor must not be able to burn the governor's next nonce"
+        );
+    }
+
     #[test]
     fn gate_rejects_profile_mismatch_without_advancing() {
         let mut gate = gate();
--- a/crates/kirra-release-token/tests/restart_trust_epoch.rs
+++ b/crates/kirra-release-token/tests/restart_trust_epoch.rs
@@ -1,0 +1,188 @@
+//! #1214 — what a consumer restart does to replay protection.
+//!
+//! The intended assertion for this test was:
+//!
+//!     mint a token under epoch A → restart the consumer →
+//!     present the epoch-A token → REJECTED
+//!
+//! **The protocol cannot make that assertion true today**, and this file pins
+//! what actually happens instead. Writing the aspirational assertion and marking
+//! it ignored would be worse than useless: it would read as a covered case.
+//!
+//! The gate's three watermarks — `last_released_sequence`, `last_released_nonce`
+//! and `latest_perception_frame` — are in-memory. A restarted consumer therefore
+//! accepts any sequence, any nonce and any perception frame. The ONLY thing that
+//! refuses a pre-restart token is its own wall-clock lifetime, so the exposure
+//! window is exactly `maximum_lifetime_ms` wide and closes by expiry rather than
+//! by any restart-aware rule.
+//!
+//! In the shipped configuration that lifetime is 200 ms
+//! (`ROS_BOUND_RELEASE_LIFETIME_MS`) and a real process restart takes far
+//! longer, so this is not exploitable as configured. That is a **timing
+//! accident, not a designed invariant** — it depends on restart duration
+//! exceeding a configurable token lifetime, and a backward wall-clock step
+//! during the restart widens it further.
+//!
+//! The structural fix (a boot/session epoch bound into evidence identity and
+//! release tokens, so pre-restart tokens are invalid by construction) is
+//! tracked separately. See the issue cited in the tests below.
+
+use ed25519_dalek::SigningKey;
+use kirra_release_token::ros_bound_command::{
+    issue_ros_bound_command_release, RosBoundCommandGate, RosBoundCommandPayload,
+    RosBoundCommandRefusal,
+};
+
+const MAXIMUM_LIFETIME_MS: u64 = 200;
+const PROFILE: [u8; 32] = [0x11; 32];
+
+fn key() -> SigningKey {
+    SigningKey::from_bytes(&[42u8; 32])
+}
+
+fn payload() -> RosBoundCommandPayload {
+    RosBoundCommandPayload {
+        sequence: 10,
+        nonce: 9001,
+        issued_at_ms: 10_000,
+        expires_at_ms: 10_000 + MAXIMUM_LIFETIME_MS,
+        linear_mps: 0.75,
+        angular_rad_s: -0.25,
+        scan_sequence: 77,
+        camera_sequence: Some(88),
+        tracker_generation: 3,
+        profile_digest: PROFILE,
+        evidence_digest: [0x22; 32],
+        proposal_digest: [0x33; 32],
+    }
+}
+
+/// A restarted consumer is a NEW gate: same key, same pinned profile, no
+/// watermarks. This is what "restart" means throughout this file.
+fn restarted_consumer() -> RosBoundCommandGate {
+    RosBoundCommandGate::new(key().verifying_key(), PROFILE, MAXIMUM_LIFETIME_MS)
+}
+
+/// Documents the gap: a pre-restart token IS accepted after a restart while it
+/// remains inside its own wall-clock lifetime.
+///
+/// This test asserts behaviour the project does not want. It exists so the gap
+/// is visible in the test suite rather than only in prose, and so that
+/// implementing epoch binding forces a deliberate edit here rather than silently
+/// leaving a stale expectation.
+#[test]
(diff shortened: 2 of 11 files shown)


justinlooney:
All four gates worked through. **Gate 2 found a real bug in my own fix** — the hold arithmetic was wrong at the boundary, in the direction that matters. Pushed through `240f3e92`.

## Gate 2 — the hold was too short by exactly the step magnitude

The reference model you asked me to pin is what surfaced it. Let `L` be the token lifetime and `δ` a backward step. A token minted at `T` expires at `T + L`; after the step the consumer's clock reads `true_time − δ`, so it accepts while

```
true_time − δ < T + L    ⟺    true_time < T + L + δ
```

Worst case `T = t₀` (minted at the step instant), so a pre-step token stays acceptable until **`t₀ + L + δ`**. My hold ran to `t₀ + L`, leaving a `δ`-wide window admitting a stale token — the exact failure the module exists to prevent, reintroduced at the edge.

Demonstrated against the running code before fixing:

```
step detected: CLOCK_STEP_DETECTED, hold_remaining=200
guard released after 200 ms of real time
STALE TOKEN STILL ACCEPTED BY THE GATE: True
needed hold >= L + delta = 350
```

It now releases at exactly 350.

**`δ` cannot be dismissed as "large steps are caught anyway."** I checked that reasoning and it fails: a large backward step does not *prevent* acceptance, it *delays* it. The token reads `FutureIssued` immediately after the step, becomes acceptable again at `t₀ + δ`, and stays acceptable for `L`. The window is shifted, not removed.

The extension is backward-only. A forward step moves the clock ahead so tokens expire sooner — fail-closed, no extension possible. I deliberately did **not** add `δ` there: it would be conservative but misleading, implying both directions carry the same risk.

Your six points are now in the module docstring, and the boundary is pinned from both sides — one millisecond early the frame is refused **and** the stale token would still have been accepted by the gate (so the refusal is load-bearing, not a coincidence of a token that had already died); at the deadline the guard releases **and** the token has genuinely expired. One ms shorter admits a stale token; one longer is unnecessary refusal.

**Restart** is documented rather than defended, with the reasoning. The hold is in memory — but so are the gate's `last_released_sequence`, `last_released_nonce` and `latest_perception_frame` (verified: no serde, no persistence on either side). A restarted consumer already accepts any sequence and any nonce. Persisting the clock hold alone would close the narrowest of four windows while leaving the wider ones open, and would make restart *look* defended when it is a full re-establishment of trust. Recorded as `EVIDENCE_BINDING.md` §4.7 with the operational consequence.

## Gate 1 — zero is now unrepresentable, not merely unwritten

You were right that the type is stronger than scattered guards. The live counter is `NonZeroU64`. `saturating_add` cannot wrap to zero today, but nothing except the type stops someone changing it to `wrapping_add` — and the original defect *was* a single initializer holding the wrong value.

Paths audited: fresh process, tracker reset, tracker recreation, state restore, rollover, fixture/default construction. Every constructor funnels through `with_options` (`new` and `Default` both route there); there is no `Deserialize`, so no state-restore path exists. Rollover saturates at `u64::MAX` — pinned by a test, since 2^64 resets is not reachable in one. The degradation at the ceiling is that resets stop reading as new epochs: a real loss, conservative, and not one where frames read as invalid evidence.

**The composed test is the one that mattered.** The defect lived in a seam between two individually correct components, and a unit test asserting "Taj emits 1" would have missed it — the number alone is not the claim; the claim is that it is a number the next component accepts. `tests/cold_start_evidence_binding.rs` drives real types across the real seam: a `PerceptionResponse` from a cold `TrackedPerceptionState` → real `PlanRequest` → real `handle_plan`. Verified against the actual defect by publishing generation-1 as zero, which reproduces the original refusal verbatim:

```
INVALID_PERCEPTION_FRAME_ID — perception tracker_generation must be greater than zero; NO MOTION
```

## Gate 3 — threat boundary stated, terminology ruled

Your statement is in as §1a, positioned where someone about to cite the binding as a control will hit it, with an explicit split of what the binding does and does not provide, and a terminology rule: *bound*, *attributed*, *non-replayable* are accurate; *verified sensor evidence*, *authenticated perception*, *proves the digest corresponds to the frame* are not, and must not be used unless a component actually recomputes or attests.

## Gate 4 — criteria corrected on the issue

Recorded on #1214: for AC4 and AC5, what was already implemented, the evidence, which tests were missing, and why no production change was needed. Both premises were false — AC4 also assumed the payload names a track, which it does not.

Also noted there that the issue's checklist was a hypothesis and testing it *was* the work: the two things that actually mattered were not on the list.

## Separate issues filed

- **#1229** — commit signing configured but silently producing unsigned commits. Framed as you described: the defect is that the config requests signing and git reports success without producing one, so a developer cannot distinguish "unavailable" from "signed". Scoped as prospective, explicitly not a history rewrite.
- **Vercel** stays out of this PR — infrastructure producing false-red before the ignored-build logic can execute.

## Merge posture

Not asking for a merge on checklist completion. Against your five conditions: (1) done — type-level, all paths audited, rollover pinned; (2) done — both boundary directions, worst-case token; (3) resolved explicitly as documented-not-defended, with reasoning; (4) done; (5) done. Condition (3) is the one worth a second opinion, since I chose documentation over persistence — the argument is that partial persistence misrepresents the restart story, but that is a judgement call.

#1214 stays open after merge until cold-start and clock-step behaviour are verified in the deployed runtime, for the reason you gave on #1212 — and with particular force here, since the defect was specifically a binding that looked present while being inert from process start.

**Verification:** 74 release-token · 152 sidecar · 3 cold-start integration · 721 Python · fmt clean · clippy clean. Every new assertion mutation-verified: reverting to the `L`-only hold fails 5 tests, extending forward steps fails 2, publishing generation-1 as zero fails the composed tests.

---

justinlooney:
Wrote the composed restart test. **The first assertion cannot be made true today**, so per your instruction I've captured it as a separate issue rather than pretending the clock hold solves it.

## What the probe found

```
epoch-A release:                              Ok
after restart, token within its window:       Ok      ← the finding
after restart, token past its window:         Err(Expired)
```

A pre-restart token **is** accepted after a consumer restart, provided it is still inside its own wall-clock lifetime. Nothing refuses it because a restart happened — the sequence and nonce that would have refused it were cleared along with everything else. Only expiry closes the window.

Your conditional was the right one to set:

> If an old token can survive a verifier or consumer restart because its signature and wall-clock lifetime remain valid, then the absence of persisted nonce and sequence state becomes the larger issue.

It can, and it does. **#1230** filed, with boot/session epoch binding (your option 1) as the preferred fix — it fits the existing shape, since the payload already carries `tracker_generation` as a perception epoch and this is the process-level analogue. I rejected option 2 for the same reason you rejected the partial version, and noted that option 3 moves the problem into key provisioning.

## Why the exposure is not reassuring even though it is small

The window is exactly `maximum_lifetime_ms` — 200 ms shipped, against a restart that takes far longer. But that is a timing accident, not an invariant, and it rests on three things the design never asserts:

- restart duration exceeding a **configurable** lifetime. `KIRRA_FRESHNESS_WINDOW_MS` widens the post-restart replay window by exactly the amount it is raised, which is invisible from its name and its documentation. Now recorded in both places.
- no backward clock step during the restart — and the #1214 guard cannot help, because its hold died with the process.
- restart never becoming fast: supervisor respawn, crash loop, container restart with a warm cache.

## The test asserts the undesired behaviour on purpose

`crates/kirra-release-token/tests/restart_trust_epoch.rs` pins what actually happens, with the aspirational assertion called out in the failure message. Writing the desired assertion and marking it ignored would have been worse than useless — it would read as a covered case. This way, implementing #1230 forces a deliberate edit rather than leaving a stale expectation.

Paired with a non-vacuity test showing the restarted gate is not simply accepting everything: it re-establishes and then enforces normally within its new epoch. That test also makes a subtle point explicit — a stale token arriving *after* the new epoch is established is refused by the new sequence watermark, not by anything that knows a restart happened. Had it arrived first, it would have been accepted.

## §4.7 rewritten

The restart model is now stated without qualification — all in-memory state lost, trust re-established from scratch, no guard claiming continuity — with what re-establishment already satisfies separated from the one requirement it does not. My earlier draft understated this by framing restart purely as "the clock hold is not persisted"; that was accurate but incomplete in a way that flattered the design.

## Against your merge conditions

| | |
|---|---|
| clock hold boundary tests green | ✅ both directions, worst-case token, tight to the millisecond |
| composed cold-start acceptance test green | ✅ real types across the real seam |
| backward-step magnitude included in the hold | ✅ `L + δ`, backward-only |
| restart threat model explicit | ✅ §4.7, stated as full loss of continuity |
| separate issue if pre-restart tokens survive state loss | ✅ **#1230** — they do |
| deployed verification before closing #1214 | pending, as agreed |

Also #1229 for the signing defect.

**Verification:** 74 release-token (+4 restart) · 152 sidecar · 3 cold-start integration · 436 robot · 285 ROS · fmt clean · clippy clean.

---

Issue #1218, opened 2026-07-28
[R2 perception] Non-blocking perception path — the scan callback still blocks on HTTP JSON


justinlooney:
Child of #1209. Audit phase 8. **Verified status: gap — confirmed exactly as the plan stated.**

## Verified current state

Taj is still reached by `POST /perception` with a JSON body. There is no latest-request-only slot, no stale-reply discard, and no request/response sequence IDs. The sidecar architecture and service boundaries are real and well-drawn; the transport underneath them is not production-shaped.

## The failure this leaves open

A scan callback that blocks on an HTTP round trip couples the sensor callback's period to a network path. Under load the callback backs up; the queue grows; and when replies do arrive there is nothing to establish that the reply belongs to the scan currently in hand. A reply for scan N applied to scan N+3 is perception evidence attributed to the wrong instant, which is exactly the failure the evidence-frame work on #1214 exists to make impossible — except that binding cannot help if the mismatch happens before the frame identity is stamped.

## Acceptance criteria

- [ ] Scan callback returns without blocking on any network or HTTP call.
- [ ] **Latest-frame slot** — a new scan overwrites a pending un-started request rather than queueing behind it. Bounded by construction, not by a large queue.
- [ ] **One in-flight job** at a time; a second is not started while the first is outstanding.
- [ ] **Stale-result rejection** — a reply whose request sequence is not the one currently outstanding is discarded, counted, and never applied.
- [ ] Request/response **sequence and generation IDs** so overwrite is detectable rather than silent.
- [ ] **Explicit deadline** per request; a breach derates via the perception cap rather than waiting.
- [ ] Discarded-reply and deadline-breach counters exposed as metrics.
- [ ] Fail-closed: a perception service that stops answering floors the cap on the deadline, not on a socket timeout.
- [ ] Migration path toward typed ROS 2 messages or iceoryx2 shared memory recorded, with the JSON path marked transitional. The iceoryx2 carrier tree already exists (`tools/iceoryx2-spike/`) — see #270 / #339 before choosing.

## Related

- #339 — rmw_iceoryx2 as ROS 2 middleware for the sensor→inference→planning chain. The long-term answer to this issue's transport half.
- #270 — iceoryx2 transport adoption epic.

## Done when

The callback returns immediately, a reply for a superseded scan is provably discarded, and a stalled perception service floors the cap on the deadline.

justinlooney:
[...]

Issue #1213, opened 2026-07-28
[R2 perception 4/8] Ackermann swept-footprint containment — corridor width is checked, the swept envelope is not


justinlooney:
Child of #1209. Audit phase 3. **Verified status: partial — doc comment only.**

## Verified current state

Minimum corridor width checks exist and fail closed: a too-narrow corridor, a malformed width and a missing width are all refused, and contradictory platform-versus-lane dimensions fail validation.

Swept-footprint checking appears **only in a doc comment in `validation.rs`**. There is no implementation. `CorridorHealth` does not appear on any wire contract.

## The failure this leaves open

The R2 is an Ackermann platform. Under steering, its body sweeps an arc wider than its static footprint, and the rear inside corner cuts inboard of the front wheel path. A corridor that admits the static rectangle can still be struck by the swept body. Checking width against a stationary footprint is the right check for a differential-drive robot turning in place and the wrong check for this one.

## Acceptance criteria

- [ ] Swept envelope computed from: current steering angle, proposed steering angle, steering **rate** limit, wheelbase, front and rear overhang, body footprint, required clearance, trajectory curvature.
- [ ] The envelope is curvature-dependent — a straight-line proposal and a full-lock proposal do not produce the same clearance requirement.
- [ ] Steering **rate** is honoured: the reachable envelope between now and the next cycle is bounded by how fast the wheels can actually turn, not by the commanded angle alone.
- [ ] Rear inside corner is modelled explicitly; a test that passes on the front axle and fails on the rear corner must fail.
- [ ] `CorridorHealth` is an explicit state on the public wire contract, not inferred from a width number.
- [ ] Station-specific minimum-width evidence recorded for the physical spaces the R2 is expected to traverse.
- [ ] Fail-closed: an unavailable steering state, a nonfinite curvature or a missing wheelbase refuses rather than falling back to the static footprint.
- [ ] Property tests: clearance requirement is monotonic in |steering angle|; a swept envelope is never smaller than the static footprint.
- [ ] Non-vacuity: a deliberately-narrowed envelope must red the containment tests.

## Done when

A proposal that fits the static footprint but not the swept envelope is refused, with a distinct deny code, and the refusal is proven by a test that passes under the old check.

justinlooney:
## Re-scope: this is a verification issue, not an implementation one

[...]

Issue #270, opened 2026-06-11
EPIC: iceoryx2 transport adoption for the QNX governor lane


justinlooney:
**iceoryx2** is now the recommended Eclipse implementation — iceoryx *classic* is in maintenance. v0.7.0 added **QNX 7.1** as a tier-3 open-source platform (tier-1 via **ekxide** commercially); **QNX 8.0** builds need `--no-default-features` (std-dependency gaps).

## Architectural significance
The C++ shim existed because iceoryx classic is C++. With a **Rust subscriber edge**, the governor hot path can be **Rust end-to-end** — judge + driver as Rust modules, **no FFI on the command path** — coherent with the QNX + Ferrocene certification stack. The FFI layer is **demoted** to the documented C/C++ integration boundary (DDS bridges, vendor stacks), not the command path.

## Architecture context
This lane is the **boundary** domain of the three-domain stack — **safety partition (QNX governor)** ↔ **boundary (the hypervisor line)** ↔ **autonomy guest (Ubuntu / Autoware + parko)**. iceoryx2 lives **in-partition**; the boundary itself is the **frozen-layout contract** (ADR-0006 Clause 2), not a transport library — the governor judges every crossing. The decided integration of all seven technologies into one narrative is **#295** (`docs/ARCHITECTURE_STACK.md`); the live cross-domain board index is **#302**. This EPIC carries `domain:boundary`.

## Children
- [x] #271 — Land `tools/qnx-rtm-harness` (FFI boundary + Rust judge + FDIT/RTM harness, corrected) — ✅ **merged #284**
- [x] #272 — Trace the RTM harness matrix to real safety-goal IDs — ✅ **merged #285** (1 genuine hit SG-001/TR-001 proxy + 6 honest NO-RTM-ID gaps + 1 control)
- [ ] #273 — iceoryx2 feature-subset spike — host side (Linux) _(depends on #271; PR #277 in review)_
- [ ] #274 — iceoryx2 on QNX 8.0 — cross-compile + feature-subset verification _(depends on #273; **hardware-blocked**)_ — WCET measurement methodology landed (#287, `KIRRA-OCCY-WCET-METH-001`); on-target measurement still blocked
- [ ] #275 — ADR: governor transport = iceoryx2 (Rust end-to-end); FFI demoted to integration boundary _(depends on #273)_
- [ ] #276 — Evaluate ekxide tier-1 / commercial support (certification path) _(owner action)_

[...]

Issue #1209, opened 2026-07-28
[EPIC] R2 safety-path gap closure — code-verified audit of the 2026-07-27 plan


justinlooney:
Tracking epic for the gaps confirmed by a code-verified audit of the 2026-07-27 R2 plan (`Robot_plan_260727`), run against head `8fc41d47`.

## Why this epic exists

The source plan states it was based on supplied evidence rather than a direct reading of the repository. Each of its eighteen phases was re-checked by searching for the specific artifact it names — type names, field names, file paths, document titles — across `crates/`, `src/`, `robot/`, `ros2_ws/`, `parko/`, `docs/` and `tests/`, then reading the hits to distinguish implementation from documentation or test scaffolding.

Result: **six phases were understated** (the capability exists), **seven gaps confirmed absent**, **four confirmed substantially complete**, and **one exists in the wrong layer**. Only the confirmed-absent and wrong-layer findings are filed below. Nothing already shipping is re-litigated.

## Children

Ordered by the revised plan. Items 1–3 are prerequisites for any hardware motion claim; 4–6 close verified gaps; 7–8 are the long-horizon items. The last three are confirmed gaps outside the headline eight.

- [ ] #1210 — **1/8** LiDAR self-filter calibration · *gap, zero implementation*
- [ ] #1211 — **2/8** R2 sensor watchdogs · *gap, all four checks absent*
- [ ] #1212 — **3/8** Move cap stabilization into the checker · *exists in Python, wrong layer*
- [ ] #1213 — **4/8** Ackermann swept-footprint containment · *doc comment only*
- [ ] #1214 — **5/8** Evidence-binding residue · *5 of 7 items already ship*
- [ ] #1215 — **6/8** Shared geometry types + joined cycle record · *types absent, recorder half present*
- [ ] #1216 — **7/8** Hardware evidence campaign · *infrastructure present, exercise not run; partly Rev-A blocked*
- [ ] #1217 — **8/8** Perception depth and data · *the scale gap; separately resourced*
- [ ] #1218 — Non-blocking perception path · *scan callback still blocks on HTTP JSON*
- [ ] #1219 — Canonical platform manifest + baseline; retire `courier` · *digest enforcement live, manifest absent*
- [ ] #1220 — Safety-case residue: standalone ODD spec + residual-risk register

## Phases the plan understated — no work filed

| Phase | Plan said | Repository shows |
|---|---|---|
[...]

PR #1231, opened 2026-07-29
feat(perception): make the scan callback non-blocking and identity-checked

justinlooney:
> Measured: the scan callback returns in **~0.4 ms against a sidecar taking 5 s**. It used to wait for all of it.

**Completes #1218.**

## The two failures

`perception_governor._on_scan` POSTed to Taj and blocked on the reply **inside the scan subscription callback**. Under rclpy's single-threaded executor that couples sensor ingest to a network round trip: while the callback waits, no other callback on the node runs and inbound scans queue behind it. A slow sidecar backs up the very stream it is being asked to judge.

Nothing established that a reply belonged to the scan in hand, either. A reply for scan N applied to scan N+3 is perception evidence attributed to the wrong instant — and as the issue notes, the #1214 evidence binding cannot help, because that mismatch happens *before* the frame identity is stamped onto anything.

## Shape

```
_on_scan   →  stamp, copy, offer to a one-entry slot, RETURN
worker     →  one at a time, does the network, deposits
timer      →  the only publisher; decides once per period
```

`async_perception` holds the decisions, pure and ROS-free — the same split `async_plan` already uses for the doer's plan cycle. **This applies a proven in-repo pattern to a second call site rather than inventing one.**

It differs where the problem differs. The doer runs on a timer and asks "may I start work?"; perception is driven by *arrival*, which the node does not control, so a scan landing mid-request has to go somewhere. It goes into a **latest-frame slot** of exactly one entry, where a newer scan displaces an older un-started one.

Overwriting is right for perception in a way it would not be for commands: the newest scan is strictly better evidence, and the older one's only remaining value would be to make the robot act on a staler view of the world. Bounded by construction, not by a queue length nobody tuned.

Displacements are **counted**, alongside discarded replies and deadline breaches, on the health topic an operator is already reading. Bounded-by-construction is a design property; how often it engages is an operational fact, and a governor dropping every second scan should be visible to whoever is asking why the cap is low.

## The deadline is the node's, not the transport's

`requests`' timeout is **per socket operation**. A server that dribbles bytes holds a worker open indefinitely without ever tripping it. Perception drives a *speed cap*, so it must be floored on a schedule the node controls — hence `deadline_ms`, judged on the node's monotonic clock, independent of whether the HTTP call has returned.

This is also why the publish timer is not an implementation detail that could be dropped by letting the worker publish directly: **a worker blocked on a socket cannot notice that it is late.** Something else has to be running.

The deadline and the staleness budget answer different questions and are kept separate — one bounds how long the node waits, the other how old perception may be. A reply that missed the deadline is still usable if its scan is fresh; refusing it would let one slow round trip permanently discard good, current evidence. The breach counter stays on the wire so the episode does not become invisible once a reply lands.

## One publishable outcome

`ABSENT`, `SUPERSEDED`, `STALE`, `FAULT` and deadline-breach all publish the MRC floor. They stay distinct because the *operator response* differs, not the verdict. `FLOOR_STATES` is written out explicitly rather than derived as "not USE", so adding a state forces a decision about which side of the line it falls on.

The publish-once watermark matters more than it looks: without it, a wedged sidecar's last good answer would be re-served every period, and **a frozen world would look exactly like a clear one**.

## Lock discipline

Decide under the lock, publish outside it. Holding a lock across a ROS publish would couple the worker's deposit to publisher latency — the coupling this issue exists to remove. The request body is likewise built *before* the lock: it must be copied off the live ROS message, but that copy is pure computation and holding the lock across it stalls the worker for no reason.

The in-flight flag and the slot share one lock deliberately: "is a job running" and "what would it run on" is one question, and splitting it is a race that starts two workers.

## Tests — 34

The non-blocking property is pinned **structurally as well as behaviourally**: a test parses the node source and follows self-calls out of `_on_scan` looking for anything that waits. That is a property of the source rather than of whichever branch a test happens to exercise, and it comes with a non-vacuity companion proving the detector fires on `_fetch`, where the blocking call legitimately lives.

Mutation-verified — each of these fails a test:

| Mutation | Result |
|---|---|
| callback performs the HTTP itself (**the original bug**) | ✅ killed |
| deadline never enforced | ✅ killed |
| publish watermark never advances | ✅ killed |
| second concurrent request allowed | ✅ killed |
| slot keeps the older scan | ✅ killed |
| republish an already-published scan | ✅ killed |

## Transport is transitional

Recorded on the module, because nothing here should read as endorsing JSON-over-HTTP for a per-scan safety signal. The migration target is a typed carrier (#270/#339). **The decisions survive that move** — one in flight, newest-scan-wins, publish-at-most-once, deadline-floors-the-cap are properties of the cycle. The request/response sequence pairing does not; it exists only because this transport cannot express identity itself.

## Note on #1213

While picking this up I verified #1213's premises and **its headline is already implemented**. The C1 swept-footprint check landed 2026-07-27, a day before that issue was filed — all four corners' chords between consecutive poses, inflated by the segment sagitta, rear corner modelled explicitly. The audit appears to have grepped `validation.rs`, where the doc comment is, and missed `containment.rs`, where the implementation is. What genuinely remains there is smaller: steering-*rate* reachability, a `CorridorHealth` wire state, and field-measured station widths. Worth someone re-scoping that issue before it is picked up as written.

**Verification:** 755 Python tests · node and config parameters agree in both directions · no dead config entries.

---

### DIFF
--- a/ros2_ws/src/kirra_safety/config/kirra_params.yaml
+++ b/ros2_ws/src/kirra_safety/config/kirra_params.yaml
@@ -112,7 +112,14 @@ perception_governor:
     scan_topic: "/scan"
     cap_topic: "/kirra/perception_speed_cap"
     health_topic: "/kirra/perception_health"
-    timeout_ms: 40
+    timeout_ms: 40               # PER-SOCKET-OPERATION budget; see deadline_ms
+    # #1218 non-blocking cycle. deadline_ms is NOT a second timeout: `requests`
+    # times out per socket operation, so a sidecar dribbling bytes never trips
+    # timeout_ms. This is the wall-clock budget the NODE enforces, after which
+    # the cap is floored regardless of what the transport is doing.
+    publish_period_ms: 100       # how often the cap is (re)decided and published
+    deadline_ms: 250             # in-flight budget before the cap is floored
+    scan_stale_ms: 500           # max age of the SCAN behind a published cap
     forward_extent_m: 8.0        # Taj geometric horizon (m)
     decel_mps2: 1.5              # assumed comfortable brake for the ACD cap
     margin_m: 0.4                # standoff distance (matches the checker's lateral margin)
--- a/ros2_ws/src/kirra_safety/kirra_safety/async_perception.py
+++ b/ros2_ws/src/kirra_safety/kirra_safety/async_perception.py
@@ -49,10 +49,28 @@ counters, no slot state beyond its own deposit. Every publish decision is made
 by the executor thread calling into this module. These functions are pure, so
 they hold no lock and can never be the thing that stalls a callback.
 
-The timer is not an implementation detail that could be dropped by having the
-worker publish directly. It is what makes the DEADLINE enforceable: a worker
-blocked on a socket cannot notice that it is late, so something else has to be
-running in order to floor the cap on time. See `deadline_breached`.
+THE TIMER IS AN INDEPENDENT LIVENESS AUTHORITY, NOT A POLLING LOOP
+------------------------------------------------------------------
+Read this before moving deadline handling anywhere else.
+
+The worker CANNOT be trusted to report that it missed its deadline, because the
+failure mode is precisely that it may never regain execution. Self-reported
+lateness is only available from a thread that is still running, which excludes
+every case the deadline exists for.
+
+So the timer is not an implementation detail that could be dropped by having the
+worker publish on completion. It is the one component that can, without
+depending on transport progress at all:
+
+  - age the last result out of usefulness,
+  - floor the cap when the deadline passes,
+  - report displacement and blockage,
+  - and surface that process-level recovery is needed.
+
+A refactor that moves deadline handling into the worker — "simpler, it already
+knows when it started" — silently removes every one of those, and does so in a
+way that passes every test written against a sidecar that eventually answers.
+The failure only appears against one that never does. See `deadline_breached`.
 
 WHY A DEADLINE IS NOT A SOCKET TIMEOUT
 --------------------------------------
--- a/ros2_ws/src/kirra_safety/kirra_safety/async_perception.py
+++ b/ros2_ws/src/kirra_safety/kirra_safety/async_perception.py
@@ -74,6 +74,30 @@ cannot presently justify a non-zero speed cap, so it publishes zero. They are
 kept as distinct reasons because the operator response differs, not because the
 verdict does.
 
+WHAT THIS DOES NOT FIX — a claim boundary worth stating
+-------------------------------------------------------
+This makes the ROS EXECUTOR recoverable. It does not make a blocked socket
+recoverable. Those are different claims and conflating them would be the most
+consequential misreading of this module.
+
+If a worker never returns — a socket that neither completes nor times out — the
+one-in-flight bound means no replacement is ever started, and the newest scan
+sits in the slot unprocessed for as long as that lasts. The cap stays floored,
+so the outcome is SAFE; but the node does not recover on its own, and
+availability is lost indefinitely.
+
+That bound is not a defect to remove. It is what stops a sick sidecar exhausting
+threads, and abandoning a worker whose socket may still be live would trade a
+bounded availability loss for an unbounded resource one.
+
+Recovery therefore belongs to the RUNTIME, not to this module, and the options
+are process-level rather than thread-level: supervision that restarts the node,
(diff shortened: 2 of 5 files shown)


justinlooney:
Worked through the review. **One of your points found a claim I was overstating**, and it's the most important thing here.

## The availability limitation — I was overclaiming

You were right to separate these:

> It makes the ROS executor recoverable; those are different claims.

Probed against the running code with a socket that never completes and never times out:

```
worker calls made:       1
in_flight still set:     True
newest pending seq:      6      ← never processed
deadline breaches:       6
all caps floored:        True
distinct health reasons: ['TAJ_DEADLINE']
```

So: **safe, and not recovered.** Exactly one request is ever sent, and the newest scan waits in the slot indefinitely. My PR body said the callback returns fast and left the reader to infer the rest, which is the kind of omission that reads as a promise.

The module now carries an explicit claim boundary, and I've argued there why the one-in-flight bound is **not a defect to remove**: abandoning a worker whose socket may still be live trades a bounded availability loss for an unbounded resource one. Recovery is a runtime concern and process-level rather than thread-level — supervision, sidecar circuit-breaking, a replaceable worker *process*, or transport cancellation. None is implemented, and the test is named `test_a_wedged_worker_blocks_every_later_scan_which_is_an_availability_limit` so a future change adding recovery has to come here and edit it rather than quietly satisfying it.

## Deadline × freshness matrix

Pinned as four corners rather than prose, because you're right about which "simplification" is coming:

| | fresh scan | stale scan |
|---|---|---|
| **before deadline** | `USE` | `STALE` — fast transport is not evidence of fresh perception |
| **after deadline** | `USE` — *the case that must not be removed* | `STALE` |

Plus `test_freshness_is_judged_at_publish_time_not_at_reply_time`: the identical held result ages out where it sits with no new information arriving. That's the property stopping a wedged sidecar's last good answer being served forever.

## Race determinism

You asked about completion-vs-arrival. `_run_request` calls `_maybe_issue` after depositing and `_on_scan` calls it too, so both paths can try to start the successor simultaneously. A four-thread scan flood against a fast sidecar holds `max_concurrent` at 1, with a non-vacuity assertion that the flood drove real requests rather than none.

## The split-snapshot hazard you flagged

> response metadata from frame A published alongside freshness or identity from frame B

Checked: `resolve_result(self._result, …)` and `accepted = self._result` are both inside the same `with self._lock`, and the worker's deposit (`self._result = outcome; self._in_flight_since_s = None`) takes the same lock. The result is an immutable namedtuple carrying its own `request_sequence` and `scan_received_s`, so identity and freshness travel with the response and cannot be paired across frames. Displacement touches only the slot; the in-flight request lives as a worker-local, so it has no identity to overwrite.

## Against your merge conditions

| | |
|---|---|
| `_on_scan` has no reachable blocking transport call | ✅ structurally enforced + behavioural |
| ROS publication outside worker-state locks | ✅ decide under, publish outside |
| one in-flight + one pending enforced atomically | ✅ shared lock; 4-thread flood test |
| displaced-frame accounting correct | ✅ counted, never reset, take-then-offer is not a displacement |
| response identity cannot cross generations | ✅ immutable result carries its own sequence |
| timer flooring independent of the worker | ✅ node clock; wedged-worker test |
| late replies judged separately for deadline and freshness | ✅ the matrix |
| late/obsolete worker cannot restore a cap | ✅ `test_a_late_worker_cannot_restore_a_cap_after_its_scan_has_aged_out` |
| permanent blockage documented or supervised | ✅ documented, named test, recovery options listed as unimplemented |

On your note about keeping the structural detector narrow: it follows direct `self.` calls only, asserts on a fixed blocking-name set, and has the non-vacuity companion. It does assert two method names exist (`_on_scan`, `_fetch`) — deliberate, with a message telling a renamer to update the guard, since silently passing after a rename is the failure mode that would matter.

## #1213 and the epic rule

Both done. #1213 re-scoped with provenance (`27bcdfdc`, 2026-07-27 — one day before the issue) and a residual-risk list; the genuine remaining gap is steering-*rate* reachability, which is a refinement of a working check rather than a replacement.

The rule is on #1209, with the argument for why it's worth formalising: the failure is **asymmetric in the more dangerous direction.** An audit that over-reports work is self-correcting — the engineer attempts it and finds it exists. One that under-reports is not: the plausible outcome is a second safety checker beside a working one, and for containment that means two authorities over the same envelope.

**Verification:** 765 Python tests · 44 on this cycle · mutation-verified across six mutations including the original bug.

---

Issue #1243, opened 2026-07-30
Over-ceiling speed-cap path bypasses acceleration enforcement


justinlooney:
## ⚠️ Corrections after #1242 landed (2026-07-30)

Three statements below were written before #1242 finished and are now known to be wrong. They are corrected in place; recorded here as well because one of them would send an implementer down a path we already came back from.

1. **"The K8 harness IS provable and `tan`-free" — the reasoning is wrong.** That is precisely the mistake #1242 corrected. `tan`-freeness of the *assertion* is irrelevant: **Kani fails a harness whenever an unsupported foreign call is REACHABLE**, whether or not any assertion depends on its value. K8 would call `validate_vehicle_command` over the full pipeline and so put the P6 bicycle model on its paths. Written as described below it would fail on day one with `call to foreign "C" function 'tan' is not currently supported`, no assertion violated — exactly how K3, K3b, K6 and K7 failed. K8 must carry the same `#[kani::stub]` attributes as K3/K6 (`f64::tan`, `f64::atan`, `f64::powi` — see the "Solver model" block in `verification/kani/src/proofs_kinematics.rs`). Cheap now that the model exists, but not free, and not optional.
2. **"two pin locations" — there are FOUR.** The spec doc; `ci/build_safety_case.py`'s `talisman_gate`; the `gateway::provenance` manifest/sidecar tests; and the `rustfmt gate` workflow step. Checklist in `docs/safety/CLAMP_APPLICATION_INVENTORY.md` — use it rather than rediscovering them one failure at a time, which is what happened on #1242.
3. **"twelve machine-checked properties" — fifteen**, after K3b/K6/K7. And the scope note's merge blocker is discharged: #1242 merged as PR #1244.

**Approval policy has since changed — read `docs/safety/TALISMAN_AMENDMENT_POLICY.md` before starting.** The old "reviewer approval, named" wording is retired: it could not distinguish genuine second-principal approval from the author signing their own change, and #1242 took the latter path. The control is now binary — either a **named second-principal approval recorded before merge** (§2.1), or a **formally recorded exception** carrying five mandatory fields (§2.2: who authored; why no eligible independent reviewer existed; who accepted the exception; exactly which evidence was independently machine-checked; what residual risk remains). "Approval with a caveat" is explicitly not an option. #1242 is logged as an exception in §4.

Everything else below stands, including the central judgement: **this is a safety-case decision before it is a code change.**

---

## Finding

[...]

Issue #1242, opened 2026-07-30
Kernel: P6 lateral envelope is dropped when the SPEED CAP clamps the linear axis


justinlooney:
## Finding

> When `validate_vehicle_command` clamps the linear axis at the **speed cap**, it returns `ClampLinear` alone and the steering passes through **unclamped**, even when that steering violates the P6 lateral-acceleration envelope at the enforced speed.

The accel/brake-bounded path is correct — it returns `ClampBoth` with the steering back-solved for the enforced linear. Only the speed-cap path composes wrongly.

## Reproduction

Contract: `max_steering_deg = 35`, `max_lateral_accel_mps2 = 3.5`, `wheelbase_m = 2.8`, `max_speed_mps = 5.225`, `dt = 0.1`.

```
linear_velocity_mps = 5.4, current_velocity_mps = 5.4, steering_angle_deg = 34
=> ClampLinear(5.225)
```

No steering component. But at the **enforced** 5.225 m/s, `a_lat = 5.225² · tan(34°) / 2.8 = 6.58 m/s²` — **1.9× the 3.5 m/s² envelope**. P6 at that speed permits only 19.75°.

Reproduced across the demand range, and identical whether `current_steering_angle_deg` equals the demand (rate-neutral) or 0 (full slew demand), so it is not a P5b interaction:

| steering demand | result | implied `a_lat` at enforced 5.225 m/s |
|---|---|---|
| 24° | `ClampLinear(5.225)` | 4.34 |
| 30° | `ClampLinear(5.225)` | 5.63 |
| 34° | `ClampLinear(5.225)` | 6.58 |

Contrast the accel-bounded path with the same contract (`max_speed` raised so the cap does not bind):

```
current = 5.0, want = 10.0, steering = 30  =>  ClampBoth { linear: 5.5, steering: 17.9506 }
```

`17.9506°` is exactly `atan(3.5 · 2.8 / 5.5²)`. So the composition logic exists and is correct — it just is not applied on the speed-cap branch.

## Why this matters

`ClampBoth`'s own documentation states the intended behaviour:

> `steering` is back-solved for the ENFORCED `linear` (the speed the actuator actually applies), so the resulting `(linear, steering)` pair honors BOTH the longitudinal ceiling and the lateral-accel envelope.

That is precisely this case, and it does not happen.

Impact differs by enforcement point:

[...]

PR #1244, opened 2026-07-30
fix(kernel): steering priorities always run past the speed ceiling (#1242)

justinlooney:
## ⚠️ Merge blocker: named reviewer approval for the talisman re-pin

This PR changes the **frozen kinematics talisman** and re-pins its blob hash. `docs/safety/TALISMAN_CHANGE_PLAN_1242.md` §6 requires a **named reviewer** recorded against the re-pin, and `docs/CAPTURE_PIPELINE_SPEC.md` currently says `Reviewer approval: **PENDING — must be recorded here before merge.**`

**Green CI is necessary but not sufficient here.** All 32 lanes are green on `fdeff6d1`, including the blocking Kani and Mutation gates. Please record the approving reviewer in the re-pin note before merging.

**Highest-value thing to scrutinise:** axiom **A3** in `verification/kani/src/proofs_kinematics.rs`. See "The proof model" below — it is where a wrong statement would do the most damage and be hardest to spot.

---

## The defect

Priority 2 (the effective-speed ceiling) `return`ed `ClampLinear` directly, skipping P5a (rack limit), P5b (slew) and P6 (lateral envelope) entirely. A command over the ceiling had its **steering demand executed unchecked**. Measured before the fix:

| Case | Result |
|---|---|
| 200° demand, 35° rack | executed at 200° — physically unachievable by the mechanism |
| 24° at the capped 5.225 m/s | 4.34 m/s² against a 3.5 envelope |
| any demand at a 35 m/s ceiling | envelope permits ~**0.46°**; everything passed |

## The fix

Priority 2 now records its correction into `v`/`v_clamped` instead of returning, so the steering priorities always run and the terminal `match (v_clamped, delta_clamped)` is the **only executable exit**.

> **The governing invariant:** priority decisions may accumulate restrictions, but no intermediate priority may finalize an executable command.

Observable change, deliberately minimal: these commands now return `ClampBoth { linear, steering }` instead of `ClampLinear`. **The linear magnitude is unchanged** — still exactly the ceiling.

## Why option B, and the correction that decided it

I first implemented the full fall-through (letting P3/4 also run) and recommended it on the basis that three tests needed updating. **That was under-informed.** Full fall-through makes the returned linear the *tighter* of `{ceiling, accel bound}`, which falsifies Kani **K3** — *"ClampLinear must clamp to exactly max_speed"* — one of the machine-checked properties `GOVERNOR_INTEGRITY_EVIDENCE.md` cites as proved. Amending a proved property does not belong inside a lateral-envelope fix, so this PR keeps P3/4 skipped when the ceiling binds.

The branch history keeps the red option-A commits (`ec5e94a6`, `479ccb6a`) rather than squashing them away — they are where the K3 conflict was found.

## The proof model (`297f78c4`) — and a claim retracted

**An earlier revision of this description said "K3 is intact by construction". That was wrong**, and the retraction is the most important thing on this page after the merge blocker.

K3 broke. Not on its assertion — on `tan` **reachability**:

```
Failed Checks: call to foreign "C" function `tan` is not currently supported by Kani
  Location: crates/kirra-core/src/kinematics_contract.rs:681:52
```

Kani fails a harness whenever an unsupported foreign call is *reachable*, regardless of whether any assertion depends on its value. K1–K5 were written when P2 returned early, so over-ceiling commands never reached the P6 bicycle model and the documented `tan`/`atan` exclusion cost nothing. Making P2 accumulate put P6 on those paths. K3, K3b, K6 and K7 all failed. **The related claim that K6/K7 are "`tan`-free" was also wrong** — their assertions are, their paths are not.

**General lesson for talisman work:** "does my new property reach a construct CBMC cannot model?" is the wrong question. The right one is "does this change put such a construct on any path an *existing* harness quantifies over?"

### Resolution: model the transcendentals, in the proof crate only

| Call | Model | Kind |
|---|---|---|
| `f64::tan` | nondet, assumed finite (A1) | over-approximation |
| `f64::atan` | nondet, assumed finite and `\|r\| <= pi/2` (A2), plus A3 against the recorded `tan` pair | over-approximation |
| `f64::powi` | exact squaring, `assert!` pinning exponent 2 | exact |

Nondeterministic stubs mean a discharged proof holds for **every** implementation meeting the postconditions — stronger than a proof about one libm. **The talisman is not modified for the prover**; `powi` could have been respelled `v * v` in the kernel (measured to work) but that puts a tooling-driven line in the frozen blob, so the bit-identity is measured instead by `crates/kirra-core/tests/powi_square_bit_identity.rs`.

`powi` mattered more than it looks: it lowers to the **uninterpreted** `__builtin_powi`, so `v2` is an arbitrary double and the P6 entry guard `v2 > 1e-6` is undecidable. That confounded a diagnostic harness which pinned the ceiling far below the guard and *still* failed — it read as a refutation of the reachability diagnosis and was not one.

### What the reviewer should check, in priority order

**A3** — `|y| <= |tan x| ∧ |x| <= pi/2 ⟹ |atan y| <= |x|` — is the one carrying mathematical substance. A1 and A2 merely bound the model. Three specific checks:

1. **Direction.** On the P6 path, P-RACK largely *reduces to* A3 rather than being derived from it. The proof still establishes the surrounding program logic — that P6 fires only when `|tan δ| > max_safe_tan` so A3's hypothesis genuinely holds, the radians→degrees conversion, the `* delta.signum()` sign restoration, P5a upstream, and all non-P6 paths — but the nonlinear rack-bound step itself is assumed.
2. **Domain.** The `<=` at `pi/2` is not an over-assumption: there, A2 alone supplies the bound, and `FRAC_PI_2` as a double is strictly below the mathematical endpoint. **Known model-local weakness:** `stub_tan` records only when `x.is_finite()`, so a non-finite-`x` call leaves the previous pair in place and a later `atan` could be conditioned on a stale record. Harmless under the actual caller contract (P0 rejects non-finite fields; exactly one `tan`/`atan` pair per invocation) — but that is a property of the caller, not of the stub.
3. **Coupling.** Closed by construction: the recorded pair, the hypothesis and the conclusion are all magnitudes, which is the correct form for an odd increasing function, so signs cannot open a loophole. `r`'s own sign is unconstrained and harmless — P6 imposes it via `* delta.signum()`. The non-strict hypothesis safely covers P6's strict entry condition (at equality the conclusion holds with equality).

**Scope the risk correctly.** A3 is essential only to **K7**, which is deferred and has no verdict. K3 and K6 assert on the longitudinal axis and are immune. K3b's antecedent is value-dependent, so an unsound A3 could in principle make it vacuous — **measured, and it does not**: `kani::cover!` reports both "antecedent satisfiable" and "antecedent reachable with the `ClampBoth` conclusion" as SATISFIED. (Measured locally; the cover properties are not committed — see "Follow-ups".)

## Consequence recorded, not hidden

The accel limit is still not applied to over-ceiling commands: `ClampLinear(35.0)` from 5.0 m/s over 0.1 s implies **300 m/s²** against a 2.5 limit. Same early-return defect class, different protected property, and it needs a K3 restatement — so it is filed as **#1243** with its own evidence rather than bundled here.

(Figure corrected during review: earlier notes said ~450 m/s². That is the *request's* implied acceleration and is never emitted; 300 is the executed command's.)

## Another correction made along the way

My claim that a naive fall-through would push `v` **above** the ceiling was wrong — both P3/4 branches already end in `.clamp(-effective_max_speed, …)`. The `.clamp()` was on a continuation line elided by the grep I surveyed with.

Also: **P-COMPOSE is not directly expressible in Kani** — it cannot assert which `return` executed. What is expressible is its observable shadow: every executable return must satisfy every bound the pipeline should have applied. K6 ∧ K7 falsify the old control flow without mentioning control flow.

## Evidence

**Kani — 13/13 per-PR harnesses green** (K3 22 s, K3b 32 s, K6 45 s). Property count 12 → 15.

- **K6 (P-CAP)** and **K7 (P-RACK)** are new; K7 was red pre-fix.
- **K7 is behind `deep-proofs` (weekly), and its demotion is provisional — not the same as R2's.** R2 was *measured* to exceed a budget. K7 was stopped at 23 min of CBMC time with **no verdict**, so no budget is known to suffice. If the weekly lane does not converge either, the honest response is to demote it out of the proof tier rather than leave a harness that never finishes and reads as coverage. Stated that way in the config and both workflows.
- K7's blocking per-PR gate is a new **306,180-point exhaustive grid** over the executable return (ceiling × ODD cap × commanded speed × current speed × steering demand × current steering × wheelbase × lateral cap), red against the pre-fix kernel on the 200° demand. The grid and the proofs are blind in **opposite** directions and neither covers the other.

**Mutation gate — 19 mutants, 18 caught, 1 unviable, 0 missed.** Four survived the pre-existing suite:

- Three killed by exact tolerance-boundary tests (`crates/kirra-core/tests/rate_limit_epsilon_boundary.rs`). They had survived because no test had ever landed *on* the `± 1e-9` boundary — reaching it needs `dt = 1.0` so `speed_delta.abs() / dt` is an exact identity. Each has a one-ULP-further-out companion; each verified by hand-applying its mutation.
- One is a **true** equivalence: in `effective_max_speed * signum(...)`, P0 has already denied non-finite fields so signum returns only `±1.0`, and `x * ±1.0` and `x / ±1.0` are exact IEEE results. Excluded with both premises asserted by a test, not argued in prose. It is the repo's first exclusion anchored on `line:col` — `validate_vehicle_command` has eleven `*` sites sharing one description, so the usual pattern would have excluded 22 mutants including ones the suite kills.

**Talisman re-pin: `ed00f4da` → `6a61b74f`.** (An intermediate `bbfe014b` was superseded when the P3/4 guard was reshaped from a wrapped block to two guarded conditions. An earlier revision of this description quoted `bbfe014b` and said there were *two* pin locations — both wrong.) **FOUR locations must agree**, now listed as a checklist in `CLAMP_APPLICATION_INVENTORY.md` so the next re-pin is not an excavation: the spec doc; `ci/build_safety_case.py`'s `talisman_gate`; the `gateway::provenance` manifest/sidecar tests; and the `rustfmt gate` workflow step, which had hardcoded the old prefix in its grep and is now anchored on the path instead.

Also: `speed_cap_lateral_envelope.rs` (4-angle sweep, red pre-fix, plus an accel-branch non-vacuity control); three tests edited for option A restored to their originals; two actuator-middleware tests updated to assert `ClampBoth` with reasoning inline; quality ratchet 1621 → 1634 for those comments, panic budget unchanged. **Zero ignored tests remain.**

## Follow-ups deliberately not in this PR

- **#1243** — the over-ceiling path bypassing acceleration enforcement.
- The K3b `kani::cover!` non-vacuity checks were measured locally but not committed, to keep the reviewed diff stable. They are four lines and arguably belong in the harness permanently, since a later change could make K3b vacuous silently — but that is a decision for after this review.

## Companion documents

`CLAMP_APPLICATION_INVENTORY.md` (blast radius — four mechanical apply sites, no caller changes needed), `TALISMAN_CHANGE_PLAN_1242.md` (Kani-expressibility assessment first, which is what shaped the fix; its Step 0 addendum records the `tan`-reachability finding), and `MUTATION_BASELINE.md` §8.

---


justinlooney:
## 🔴 Blocker — Kani says K3 is **not** preserved by option B

The `Kani proofs (EP-15, blocking)` lane fails on `ebbb7356`:

```
Verification failed for - proofs_kinematics::k7_executable_return_respects_the_rack_limit
Verification failed for - proofs_kinematics::k6_executable_return_respects_the_speed_ceiling
Verification failed for - proofs_kinematics::proofs::k3_speed_ceiling_clamp_exact
Complete - 10 successfully verified harnesses, 3 failures, 13 total.
```

**I need to correct the PR description.** It claims *"K3 is intact by construction and its mirror is green."* The mirror **is** green — but the mirror is not the proof. Symbolically, K3 fails.

That matters beyond a red lane, because **"B preserves K3" was the entire reason B was chosen over A.** If both options perturb K3, the A-vs-B decision has to be re-made on different grounds.

### Working hypothesis, not yet confirmed

K3 is *"SG1 P2 speed-ceiling clamp exact (magnitude = ceiling, direction preserved, ODD-cap min honored)"*. Option B does not change the ceiling **magnitude** — that part of the claim holds and the mirror exercises it. But it does change the returned **variant**: an over-ceiling command that also trips a steering bound now returns `ClampBoth { linear, steering }` rather than `ClampLinear(ceiling)`. If K3 asserts the variant, it fails for every such input — and the concrete mirror missed it because its fixtures use zero steering.

That would make this the same lesson as #1242's own fixtures, one level up: **a fixture that does not vary the other axis cannot see a change in the other axis.**

K6/K7 failing needs separate diagnosis; I have not yet read their counterexamples, and I am not going to guess at them here.

### Why I did not catch this before pushing

Kani is not installable in my environment (`kani NOT installed`), so I verified against the concrete mirror tier only and said so in the commit. That was the right disclosure but the wrong confidence: I wrote *"intact by construction"* when what I had was *"intact on the fixtures I could run."* The plan's own §5 lists "existing K1–K5 green" as required closure evidence precisely because the mirror is not a substitute — and this is the case that shows why.

### Not blockers

The seven `Vercel – kirra-runtime-sdk-*` failures are account-level deployment rate limiting (`api-deployments-free-per-day`, retry in 24h). Unrelated to this diff, no action from me.

### Next

Pulling the K3, K6 and K7 counterexamples and diagnosing each. If K3 genuinely cannot be preserved by any fix shape, then the choice is not A-vs-B but *which* K3 restatement to take to review — and that is a safety-case decision, not something to resolve inside this PR.

Merge remains blocked on this **and** on the named reviewer approval for the talisman re-pin, still `PENDING` in `docs/CAPTURE_PIPELINE_SPEC.md`.

---

justinlooney:
## K3 root cause — confirmed from source

K3 asserts the **variant**, not only the magnitude:

```rust
match validate_vehicle_command(&cmd, &contract) {
    EnforceAction::ClampLinear(v) => {
        assert_eq!(v.abs(), max, "clamped magnitude is exactly the ceiling");
        assert_eq!(v.signum(), cmd.linear_velocity_mps.signum(), "direction preserved");
    }
    other => panic!("over-ceiling must ClampLinear, got {other:?}"),
}
```

Under option B, an over-ceiling command that also trips a steering bound returns `ClampBoth`, hits the `other` arm, and panics. **Both substantive claims still hold** — the linear magnitude is exactly the ceiling and the direction is preserved; `ClampBoth` carries the identical linear value. Only *"must be `ClampLinear`"* is false.

So the required K3 amendment is smaller than I feared, and narrower than option A's would have been:

| | K3 amendment needed |
|---|---|
| **A** (full fall-through) | **substantive** — the ceiling magnitude itself changes to the tighter of {ceiling, accel bound} |
| **B** (this PR) | **minimal** — variant-agnostic: assert the *linear component* of the returned action, whichever variant carries it |

That preserves SG1's content exactly. A steering correction accompanying the same linear value is strictly *more* enforcement, not less. **The reason for preferring B still holds** — my error was claiming "no amendment at all" rather than "a smaller one."

## K6 / K7 — hypothesis, not yet confirmed

I could not read their counterexamples: `proofs_kinematics` runs first in the lane, so those blocks are near the head of the log and the log API only exposes a tail.

My hypothesis is that **K6/K7 reach the P6 `tan` path**, which `GOVERNOR_INTEGRITY_EVIDENCE.md` §2 explicitly excludes as unprovable (transcendentals). K1–K5 avoid P6 by construction; K6/K7 constrain nothing, so CBMC must model `tan`. If so, this has a consequence worth stating plainly:

> **The fix moves a previously-provable input class into the unprovable P6 region.** Before, an over-ceiling command returned before P6 ever ran, so K3 was provable over that class. Now it reaches P6, so any proof over that class inherits the transcendental exclusion.

That is a real, non-obvious cost of routing every command through the full pipeline, and it belongs in the safety case rather than being quietly absorbed. It may also mean K3's amendment needs to *narrow its input domain* (assume no steering correction is triggered), not just relax the variant.

**I am flagging this as a hypothesis.** Confirming it needs the counterexample blocks — either the raw log or a re-run with the kinematics harnesses isolated. I will not restate a proved property on the strength of a guess.

## Not blockers

- Seven `Vercel – *` failures: account-level deployment rate limiting (`api-deployments-free-per-day`), unrelated to this diff.
- `R2CP host bridge (PTY loop …)`: the known intermittent — its own dedicated lane passed on the same commit, and this diff cannot reach `kirra-r2cp`.

---

justinlooney:
## Mutation gate — 8 missed, and 6 of them are an artifact of how I wrote the guard

```
41 mutants tested in 2m: 8 missed, 32 caught, 1 unviable
```

Split by cause:

| Lines | What | Mine? |
|---|---|---|
| 542:55 `>` → `>=`, 544:33 `*` → `/` | the ceiling comparison and its `signum` multiply | **yes** — genuinely new code, genuinely untested |
| 616–629 (six mutants, `>`/`>=`, `*`/`/`, `+`/`-`) | the accel/brake arithmetic | **no** — semantically untouched |

The six are in scope only because I wrapped the accel/brake block in `if !ceiling_bound { … }`, which **re-indented ~25 lines**. `cargo-mutants --in-diff` works on changed lines, so re-indentation pulled a block I did not semantically change into the gate — exposing pre-existing coverage gaps in the accel arithmetic that no previous diff had ever put in scope.

Two honest readings, and both are true:

- those gaps are **real** — the accel/brake bound's arithmetic has no test that discriminates `*` from `/`, and finding that is a genuine (if accidental) win;
- but fixing them is **not this PR's subject**, and the re-indentation that surfaced them is avoidable.

### Recommended change to the guard

Rather than wrapping the block, narrow the guard to the two assignment sites:

```rust
if speeding_up {
    if !ceiling_bound && implied_rate_abs > contract.max_accel_mps2 + 1e-9 {
} else if !ceiling_bound && implied_rate_abs > contract.max_brake_mps2 + 1e-9 {
```

Same semantics, but it touches **2 lines instead of re-indenting 25**. For talisman work that matters twice over: it shrinks the mutation scope back to my actual change, and it shrinks the diff surface the blob re-pin has to certify — which is exactly what a reviewer of a frozen-file change wants to see.

The remaining two mutants (542/544) then need real tests: a boundary case at `|v| == effective_max_speed` exactly, and one where `signum` multiplication is distinguishable from division (i.e. a ceiling `!= 1.0`).

### Holding off

I have **not** made this change. The PR is already blocked on the K3/K6/K7 question, and that is a safety-case decision which may alter the code shape — if K3's amendment has to narrow its input domain rather than just relax the variant, the guard may need to look different anyway. Re-pinning the talisman a fifth time for a shape that might change again is churn, not progress.

Sequence I would suggest: settle the K3 restatement first, then apply the narrowed guard and the two boundary tests in one pass, then re-pin once.

---

justinlooney:
## The K3 restatement did not fix it — and that is the useful result

```
Verification failed for - k7_executable_return_respects_the_rack_limit
Verification failed for - k6_executable_return_respects_the_speed_ceiling
Verification failed for - k3b_composition_reports_both_axes      <- brand new
Verification failed for - k3_speed_ceiling_clamp_exact           <- STILL fails
Complete - 10 successfully verified harnesses, 4 failures, 14 total.
```

**My diagnosis was wrong.** I attributed K3's failure to its `ClampLinear` variant assertion. That assertion is gone — K3 now reads the linear component out of either variant — and K3 fails anyway. The variant coupling was real and worth removing, but it was not the cause.

### What the evidence now says

Look at which harnesses fail versus pass:

| | Reaches P6? | Result |
|---|---|---|
| K3, K3b, K6, K7 | **yes** — over-ceiling commands now fall through to the steering priorities | all 4 **fail** |
| K1, K2, K4, K5 | no — deny early, or sit in Degraded regions | all pass |
| lease L1–L4, RSS R1–R3 | n/a | all pass |

The common factor across every failure is **reaching the P6 `tan` path**, which `GOVERNOR_INTEGRITY_EVIDENCE.md` §2 already records as outside what CBMC can verify. K3 passed before this PR precisely because the Priority-2 early return meant over-ceiling commands never reached P6. The fix removed that early return — so K3's own input class now lands in the transcendental region.

This promotes my earlier hypothesis from speculation to the best-supported explanation. It is still not a counterexample-level proof, and I am labelling it accordingly.

### The consequence, which is a safety-case decision

> The #1242 fix moves a previously-provable input class into the region the proofs cannot cover. Always running P6 is what makes the lateral envelope enforceable — and it is also what costs K3 its symbolic coverage of over-ceiling commands.

That is a genuine trade, not a defect, and it is not mine to settle. Three options as I see them:

1. **Accept and document** — K3's symbolic scope narrows to inputs that do not reach P6; over-ceiling commands move to the exhaustive concrete grid, the established R2 pattern for solver-hard instances. Honest, but it reduces what the safety case proves symbolically.
2. **`kani::assume` the harnesses away from P6** — restores green, but this is exactly the input-domain narrowing that was ruled out: it deletes the newly-exposed case from the proof rather than confronting it.
3. **Abstract `tan` for the proofs** — an uninterpreted function with monotonicity/range bounds. Most principled, most work, and would extend what K1–K5 can cover generally rather than just unblocking this PR.

I lean toward 1 for this PR with 3 filed as follow-up, since 3 is a verification-infrastructure change that shouldn't ride on a lateral-envelope fix. But it changes what the safety case claims, so it needs an explicit decision.

**No further code changes from me until that is settled.** I have now been wrong once about this failure's cause; I am not going to guess a second time and re-pin the talisman behind it.

---

justinlooney:
## Both technical blockers resolved — and one earlier conclusion here was wrong

### The Kani failures were `tan` reachability after all

I earlier reported the reachability hypothesis as dead, on the strength of the `k3_diag` harness failing. That conclusion was wrong, and the retraction matters because the diagnostic is still readable in the history. The actual failure, from the job log rather than inference:

```
Failed Checks: call to foreign "C" function `tan` is not currently supported by Kani
  File: .../library/std/src/sys/cmath.rs, line 20, in std::f64::<impl f64>::tan
  Location: crates/kirra-core/src/kinematics_contract.rs:681:52
VERIFICATION:- FAILED
```

No assertion was violated. Kani fails a harness whenever an unsupported foreign call is **reachable**, and making Priority 2 accumulate — the point of this change — put P6 on the paths K3 quantifies over. K3 had passed for the life of the proof set and broke on a construct none of its assertions mention.

`k3_diag` was a broken instrument. It pinned the ceiling at 0.0005 m/s so `v2 = 2.5e-7`, two and a half orders below the P6 entry guard, and still failed — because `v.powi(2)` lowers to `llvm.powi.f64`, which CBMC models as the **uninterpreted** builtin `__builtin_powi`. `v2` is therefore an arbitrary double and `v2 > 1e-6` is undecidable, so `tan` is reachable on every path that reaches the guard. Model `powi` and the diagnostic passes.

**General lesson for talisman work, worth carrying beyond this PR:** "does my new property reach a construct CBMC cannot model?" is the wrong question. The right one is "does this change put such a construct on any path an *existing* harness quantifies over?"

### Resolution — model the transcendentals, don't narrow the harnesses

All of it lives in `verification/kani/`. **The talisman is not modified for the prover.**

| Call | Model | Kind |
|---|---|---|
| `f64::tan` | nondet, assumed finite (A1) | over-approximation |
| `f64::atan` | nondet, assumed finite and `\|r\| <= pi/2` (A2), plus A3 against the recorded `tan` pair | over-approximation |
| `f64::powi` | exact squaring, `assert!` pinning exponent 2 | exact |

A1/A2/A3 are stated as individual theorems about the real functions so they can be checked one at a time. A3 — `|y| <= |tan x| ∧ |x| <= pi/2 ⟹ |atan y| <= |x|` — is the one P-RACK needs and it relates the two calls, so the model is a matched pair rather than two independent stubs. Because the stubs are nondeterministic, a discharged proof holds for **every** implementation meeting the postconditions, which is stronger than a proof about one libm.

`powi` could instead have been respelled `v * v` in the kernel — measured to work — but that puts a tooling-driven line in the frozen blob. The bit-identity the stub relies on is measured instead, across subnormals, exponent extremes, overflow/flush-to-zero boundaries and the operating speed range.

**Results:** 13/13 per-PR harnesses green. K3 22 s, K3b 32 s, K6 45 s. K7 alone was stopped at 23 min of CBMC time with no verdict and moves behind `deep-proofs`. Its per-PR gate is a new 306,180-point exhaustive grid over the executable return (ceiling × ODD cap × commanded speed × current speed × steering demand × current steering × wheelbase × lateral cap, asserting P-CAP and P-RACK).

**K7's demotion is not the same as R2's, and the docs say so rather than flattening it.** R2 was measured to exceed a budget. K7 has no measured budget at all — if the weekly lane does not converge either, the honest move is to demote it out of the proof tier rather than leave a harness that never finishes and reads as coverage.

### Mutation gate: 19 mutants, 18 caught, 1 unviable, 0 missed

Three of the four survivors were killable and are killed. They had survived not because the branch was untested but because no test had ever landed **on** the tolerance boundary:

| Mutant | Killed by |
|---|---|
| `:621` `>` → `>=` | implied rate EXACTLY `max_accel + 1e-9` |
| `:627` `>` → `>=` | implied rate EXACTLY `max_brake + 1e-9` |
| `:627` `+` → `-` | implied rate EXACTLY `max_brake` — inside `+1e-9`, outside `−1e-9`, so it separates the two spellings where the boundary test above cannot |

`delta_time_s = 1.0` is what makes this reachable: it turns `speed_delta.abs() / dt` into an exact identity so the rate is controlled bit for bit. Each test carries a one-ULP-further-out companion so it pins a boundary rather than a region where nothing fires. Verified by hand-applying each mutation — every one reds exactly the test written for it.

The fourth is a **true** equivalence: in `effective_max_speed * signum(...)`, P0 has already denied every non-finite field so signum returns only `±1.0`, and `x * ±1.0` and `x / ±1.0` are exact IEEE results. Excluded, with both premises asserted by a test rather than left as prose. It is the repo's first exclusion anchored on `line:col` — `validate_vehicle_command` has eleven `*` sites sharing one description, so the usual pattern would have excluded 22 mutants including ones the suite kills today.

### Still blocking

**The talisman re-pin needs a named independent reviewer.** I cannot supply that. `6a61b74f` is pinned in all four locations and CI's pin gate is green, but approval of the amendment itself is a separate, human gate. The six items for that review are listed in `docs/safety/TALISMAN_CHANGE_PLAN_1242.md`; the Step 0 addendum there now also records the `tan`-reachability finding and the model, since a reviewer needs to check A1/A2/A3 individually.

Two non-blockers for the record: the Vercel statuses are an account-level build rate limit (`retry in 24 hours`), unrelated to this diff; and #1243 (the over-ceiling path bypassing acceleration enforcement) remains open with its own evidence requirements — this PR deliberately does not fix it.

---

PR #1440, opened 2026-08-12
Tier 3 box 3f — lineage retrieval as its own query family

justinlooney:
Closes box **3f**, the last large box in `WM_SCOPE.md`'s Tier 3 set.

`KIRRA-WM-EXPLAIN-TIER-001` asked Tier 3 for *"only the deterministic lineage CONTRACT that Tier 4 consumes"*, with two constraints attached: **bounded and paginated, with truncation visible**, and **historically correct**.

## The scoping call, and why it mattered more than symmetry

The choice was between a *walk from an answer* — follow the `EvidenceDigest` a `WorldAnswer` already carries — and a **query family of its own**. The family was chosen, and the difference is not cosmetic: a walk has no pagination story, so *"truncation visible"* would have had nothing to bite on, and no version set, so a change to which evidence comes back would reach recorded references silently.

## The rule is versioned because it can change an answer four ways

`lineage_selection` is `RuleId`'s fourth member and its **first non-reducer**. The membership test this tier applies is *"can changing this alter a derived answer"*, not *"is this a fold"* — and lineage selection alters one along four axes at once: which events are chosen, the generation bound, the order, and where a page ends.

The sharpest is the **cursor**. A recorded page-2 reference carries a cursor minted by the *old* ordering; replaying it under a new one returns a set that is neither the old page 2 nor the new one, and looks entirely ordinary. A moved version refuses rather than replays.

## The dependency set is ONE rule, and the three absences are the claim

| Rule | In? | Why |
|---|---|---|
| `lineage_selection` | yes | it decides which events, in what order, where the page ends |
| `world_current_fold` | **no** | lineage returns evidence, not folded claims |
| `entity_fold` | **no** | the subject is matched as written; no identity edges followed |
| `answer_admissibility` | **no** | an inadmissible claim is still evidence |

The `answer_admissibility` exclusion looks careless and is the important one. Lineage exists to answer *"why does this answer say what it says"*, and an event that was rejected, or expired, or that an LLM proposed and nobody confirmed, is frequently the whole explanation. So `select_lineage` deliberately does **not** filter on `claim_status`, where `world_current` does.

Each exclusion is a **tripwire**, asserted rather than written down. The moment lineage follows an identity edge, `entity_fold` can change what it says and must join the set — and the assertion goes red first. That is how `entity_fold` entered `CurrentSubject`'s set in 3h.

## Compaction degrades; it does not refuse

`read_at_generation` refuses a compacted coordinate, and must: a projection folded from a log with holes is silently *wrong* and looks exactly like a correct one. Lineage is not folded — it is the evidence itself — so a page missing a compacted span is *incomplete* rather than wrong, and the citations name exactly which generations went and under which digest. The split mirrors one the store already makes (`read_composed_at_generation` refuses, `as_of_composed` degrades).

Truncation and degradation stay **independent and both observable**: a page cut short by the caller's own limit is complete evidence, bounded; a page missing a compacted span is evidence that no longer exists. Conflating them would cry wolf on every paginated read.

## The tier boundary is where the JSON array starts

§7 records `Explain` as depending on *"derivation edges being real structure rather than a JSON array of identifiers"*. `provenance` **is** that array today, so a lineage entry carries it **verbatim and unparsed**. Walking it here would be Tier 3 inventing the structure whose absence is the reason `Explain` is Tier 4.

## The mutation battery — ten mutations against shipped code

| Mutation | Caught by |
|---|---|
| drop the generation bound | integration (load-bearing test) |
| drop the ordering | store unit + corpus control |
| inclusive cursor | integration (pagination walk) |
| `More` on a merely-full page | integration + store unit + corpus |
| version check after the store read | integration (ordering test) |
| `answer_admissibility` joins the set | 2 unit tests |
| filter to confirmed claims | integration (candidate test) |
| compaction refuses instead of degrading | integration (degradation test) |
| drop the subject filter | store unit + corpus control |
| `next_page` mints a successor past the end | integration |

**Two survived the integration tests**, both found by running the mutation rather than reading the code:

- **ordering** — `generation` is `world_events`' primary key, so SQLite returns rows in that order anyway and the unsorted rule accidentally agrees on every store the tests can build;
- **the subject filter** — the SQL pre-filters by subject, so `another_subjects_evidence_is_not_in_this_lineage` tests the *query* and not the *rule*.

Both are covered at store level (unit test + corpus control, both red on the mutation) and both are recorded in the test file's header — the tempting conclusion *"the integration tests cover these"* is false, and a later reader deleting the store-level tests as redundant would remove the only coverage there is.

The `More`-on-a-full-page mutation also survived the **first draft** of `the_final_page_yields_no_next_reference`, which asked for a 256-limit page over two events — a page that could not have been full. Sizing the bound to exactly the lineage length put the off-by-one under the test a reader would look in.

A **faithfulness control** sits under the six corpus variants: the harness with every switch off must reproduce the real rendering byte for byte. Without it the six only assert that *something* differs, and a harness wrong in some seventh way would satisfy them all while naming the wrong axis.

## The honest limits

- **Lineage follows no identity edges.** An adjudication recorded under a merged-away alias is not in the canonical subject's lineage — the same limitation `WorldView::ask` states for the subject side, and what keeps `entity_fold` legitimately out of the version set.
- **`LineageRef` is a separate type from `AnswerRef`.** `KIRRA-WM-ANSWER-IDENTITY-001` lists the pagination bound among what a reference serializes; `AnswerRef` has none because its family has no pages, and this one has no clock or staleness budget because evidence does not go stale. One struct holding the union would let a caller build a lineage reference carrying a staleness budget — meaningless, but hashed into the reference's identity, so two references for the same query would compare unequal. They share `QueryKind`, `SemanticVersions`, the refusal ordering and the reproducibility horizon.

## Verification

- `cargo test -p kirra-world-store -p kirra-world-service` — green
- `ci/check_world_semantics.py` — 5 rules versioned, pinned and recorded (`lineage_selection` v1 added to the baseline)
- `ci/check_freshness_coverage.py`, `ci/check_kirra_world_bidirectional_fence.py`, `ci/check_reexport_shims.py` — green
- `cargo fmt --all --check`, `cargo clippy --all-targets -D warnings` on both crates — clean

---

### DIFF
--- a/ci/world_semantics_baseline.json
+++ b/ci/world_semantics_baseline.json
@@ -36,6 +36,13 @@
         "corpus_digest": "c6458e88ed93ce2498a8b02f97e04bdfe3f2c666a1146ed34c081f4e56bcc33f",
         "source_pin": "93d1c69a10f75e8df9e39c9d3cee1a89733b76b8fac7f0b0a7203b25370e04ae"
       }
+    ],
+    "lineage_selection": [
+      {
+        "version": 1,
+        "corpus_digest": "750c6e83b752ea7743032696b2a57b40134e97a91492961513a61e5a1fb8bc55",
+        "source_pin": "e25f2863158f146facab6c2a135c0e6b6efb0fdaa50ee139329bf4b340baa964"
+      }
     ]
   }
 }
--- a/crates/kirra-world-service/src/answer_ref.rs
+++ b/crates/kirra-world-service/src/answer_ref.rs
@@ -98,6 +98,14 @@ pub enum QueryKind {
     /// ([`crate::read_view::TemporalLookup::semantics`]), and that set has to
     /// be derived in the one place every set is derived.
     AsOfSubject,
+    /// [`crate::lineage::LineageRef`] — one page of the evidence bearing on a
+    /// subject at a pinned generation.
+    ///
+    /// A ref family of its own, described by [`crate::lineage::LineageRef`]
+    /// rather than by [`AnswerRef`]: its parameters are a page bound rather
+    /// than a clock and a budget. That module records why the two are separate
+    /// types sharing this enum instead of one struct holding the union.
+    SubjectLineage,
 }
 
 /// **A reproducible descriptor for one answer.**
--- a/crates/kirra-world-service/src/lib.rs
+++ b/crates/kirra-world-service/src/lib.rs
@@ -32,6 +32,7 @@
 
 pub mod answer_ref;
 pub mod freshness;
+pub mod lineage;
 pub mod read_view;
 pub mod semantics;
 
--- a/crates/kirra-world-service/src/lineage.rs
+++ b/crates/kirra-world-service/src/lineage.rs
@@ -1,0 +1,477 @@
+//! **Lineage retrieval — Tier 3 box 3f, the boundary half.**
+//!
+//! `KIRRA-WM-EXPLAIN-TIER-001`:
+//!
+//! > **`Explain` stays at Tier 4. Tier 3 builds only the deterministic lineage
+//! > CONTRACT that Tier 4 consumes.**
+//!
+//! with two constraints the ruling attaches to it regardless of where the
+//! rendering ends up living:
+//!
+//! > * **Bounded and paginated, with truncation visible.** […] A lineage
+//! >   response that silently stops is worse than one that says it stopped.
+//! > * **Historically correct.** Lineage for an answer true at *T* traverses the
+//! >   evidence visible at *T*, not today's graph.
+//!
+//! The selection itself — which events, in what order, where the page ends — is
+//! [`kirra_world_store::lineage::select_lineage`], where it is versioned and
+//! corpus-pinned. This module is the *query family*: a reproducible reference,
+//! its dependency set, and the provenance handles that make each entry citable.
+//!
+//! # A reference of its own, rather than a variant of [`AnswerRef`]
+//!
+//! `KIRRA-WM-ANSWER-IDENTITY-001` says a reference serializes *"query kind,
+//! query parameters, `as_known_at`, the requested valid instant, the
+//! projection/rule version set, the snapshot coordinate, **and the pagination
+//! bound**"*. [`AnswerRef`] carries no pagination bound, because the family it
+//! describes has no pages; this one carries no `now_ms` and no staleness budget,
+//! because lineage is evidence and evidence does not go stale — it is either in
+//! the log at that coordinate or it is not.
+//!
+//! One struct holding the union would let a caller build a lineage reference
+//! carrying a staleness budget, which hashes into the reference's identity while
+//! meaning nothing — so two references describing the *same* query would compare
+//! unequal on a field neither query reads. That would falsify the one property
+//! [`AnswerRef`] exists to have. The two types share what is genuinely shared:
+//! [`QueryKind`], [`SemanticVersions`], the refusal ordering, and the
+//! irreproducibility horizon.
+//!
+//! [`AnswerRef`]: crate::answer_ref::AnswerRef
+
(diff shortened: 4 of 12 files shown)


justinlooney:
Pushed `273011f9` addressing all three review comments. The middle one was right about more than it claimed, so it's worth writing up.

## The unbounded fetch (`snapshot.rs:1003`) — fixed

Correct on mechanism, and one step short on consequence. It reads as a scaling nit; it was actually in tension with a standing ruling.

`KIRRA-WM-ANSWER-IDENTITY-001` clause 2 is **"queries are bounded"** — not a preference, but a consequence of D-9's measured 10.5 s p99 at 100 000 entities and ADR-0041 D-12's finding that *an unbounded query has no bounded cost whatever its scaling verdict*. A new query family whose fetch was unbounded by construction contradicted the ruling the box was written under, and the contradiction was **invisible because the returned answer was correctly bounded** — the page bound governed the answer while nothing governed the work.

One correction to the comment: it isn't the *"bounded and paginated"* requirement that was undermined. That one is about the answer — bounded page, truncation visible — and it was met. What was undermined is clause 2, which is about the query.

### Why it was written that way, and what was actually wrong with the reasoning

The original argument was that a query pre-applying the bound, the ordering and the page would be a second implementation of a versioned rule — the #1437 drift, *"in a place where nothing would notice."* Right about the hazard, wrong about the remedy. The lesson #1437 teaches is not *"never pre-narrow"* but *"never pre-narrow anywhere that nothing would notice."* So the narrowing now ships with the thing that notices.

The fetch applies `subject`, the as-of bound, the cursor, and `LIMIT limit + 1`. The `+ 1` is what keeps `More` detectable — and the rule's own comment already read *"one over the limit is fetched conceptually here"*, so this made actual what the rule had been assuming.

`narrowing_never_removes_what_the_rule_would_keep` runs the rule over the narrowed candidates **and** over the unnarrowed ones and asserts the selections are identical, including the boundary — swept across 75 (limit, cursor, generation) combinations, because the two bounds that can disagree are the cursor and the probe, and both are invisible on a first page that fits.

### Two tests, because one cannot catch both failures

Four mutations run against the shipped code:

| Mutation | Agreement test | Bound test |
|---|---|---|
| no probe row (`limit` not `limit + 1`) | red | red |
| as-of bound tightened to `<` | red | green |
| cursor made inclusive | red | green |
| **unbounded fetch restored** | **green** | **red** |

The last row is the one worth keeping in view. The agreement test **cannot** catch a re-widening, because the unbounded fetch is the reference it compares against — re-widening makes the two sides identical, which is exactly what it asserts. That is why `a_small_page_over_a_long_history_fetches_a_small_number_of_rows` is a separate test rather than one more assertion, and the division of labour is now demonstrated rather than argued.

## `PinnedLineage` docs (`snapshot.rs:603`) — fixed

Correct, and a real defect rather than a cosmetic one: the stray lines were `PinnedComposedRead`'s, so the *rendered* rustdoc summary for `PinnedLineage` read "The outcome of a composed generation-pinned read" and linked the wrong method. Removed.

## `render_lineage` allocation (`semantics.rs:579`) — fixed

Rewritten to push in place. Output is byte-identical, which matters here beyond tidiness: this function's output **is** the value `corpus_digest` digests, so a change that moved it by one byte would have reded the semantics baseline. `ci/check_world_semantics.py` confirms the digests are unmoved.

## Unrelated: the HA drill

`HA two-process failover drill (EP-02)` failed on the previous head. It is a two-process wall-clock drill whose own header documents phase-dependent overrun as a known flake (*"observed 5005 ms"*), and the lease variant passed that same run at 4606 ms. This branch touches `crates/kirra-world-*`, `ci/` and `docs/` only — zero files in the HA path — so the diff cannot be implicated. Watching the re-run rather than asserting it will pass.

---

PR #1441, opened 2026-08-12
3g follow-up — completeness for history and subject_summary

justinlooney:
Closes 3g's stated limit: *"`history` and `subject_summary` remain unpropagated, and a boundary query for them is the follow-up."*

One acceptance rule for both:

> A boundary answer must never report `Full` when evidence required by that family may have been removed. Conservative `Degraded` is allowed; silent loss is not.

## "Unpropagated" understated it

`WorldView` exposed exactly `ask` and `ask_as_of`. So neither family *lost* completeness — **neither had a boundary query at all.** That makes the rule a construction constraint rather than a migration: there was no existing `Full` to correct.

## The two families needed different mechanisms

Both **propagate rather than recompute** — a second verdict at the boundary is a second implementation of the rule governing whether an answer can be trusted. But they propagate *different types*, because the store computes them from different things:

| Family | Signal | Computed from |
|---|---|---|
| `history` | `Resolution` | citations overlapping a queried RANGE |
| `subject_summary` | `SummaryCoverage` | the evidence behind ONE folded row |

Coercing `SummaryCoverage::Degraded { summaries }` into `Resolution::Degraded { spans, summaries }` means inventing `spans: vec[]`, which reads as *"no compacted span bore on this"* — false, and false in the reassuring direction. `SummaryLookup` therefore carries the native type. **One source of truth per family beats one type across families.**

## `answer_admissibility` is in history's set for REFUSAL, not filtering

The subtlest membership call here, and the one most likely to be "fixed" wrongly.

History does **not** filter on admissibility: a claim `ask` declines to serve is still part of the record, and hiding it would make history lie about the past exactly as an admissibility filter on a lineage would hide the evidence an investigator came for.

It is in the set because history still **resolves** each claim's policy, so 3e's fail-closed refusal on unclassified semantics reaches this family. Changing the rule can turn a history answer into a refusal — that is the membership test, and it does not care that the rule is consulted for refusal rather than for dropping rows. **Two separate tests** pin the two halves, because as one assertion they would read as one fact.

`world_current_fold` and `entity_fold` are both out and asserted so: history returns raw confirmed claims in generation order, folding nothing and resolving no identity.

## The conservative citation rule is pinned, not sharpened

`resolution_for`'s citation check is **store-wide**, so one retained citation degrades every subject's history — including subjects the compacted span never touched. Scoping it to the queried subject would convert a conservative signal into an exact-loss detector, and an exact-loss detector that is wrong is silent.

The consequence is structural rather than stylistic: the `Full` arm is reachable **only in a store with no citations at all**, so the two history controls need two separate stores. A reader who consolidates them will find the Full arm unreachable — the intended tripwire.

## Numerical reconciliation is not evidence coverage

> Successful numerical reconciliation does not imply complete evidence coverage.

`reconciled_observation_count` and `reconciled_first_observed_ms` genuinely reconstruct their pre-compaction values from citations. `provenance_head` and `last_event_id` cannot be reconstructed at all — a citation names a *span*, not the events inside it — and for a fully compacted subject `retained` is `None` and those fields do not exist. `SummaryLookup::is_degraded` reads `coverage` and only `coverage`; there is deliberately no constructor taking a reconciliation result.

## Six mutations, and the one that separates two controls

| # | Mutation | Reds |
|---|---|---|
| 1 | history: force `Full` | `a_compacted_history_reports_degraded` |
| 2 | history: force `Degraded` | `an_uncompacted_history_reports_full` |
| 3 | summary: `is_degraded` → `false` | both summary-degraded controls |
| 4 | summary: `is_degraded` defers to reconciliation | both summary-degraded controls |
| 5 | history: swallow the `policy_for` refusal | `an_unruled_claim_refuses_the_whole_history` |
| 6 | fixture: drop the post-compaction rebuild | `reconciliation_does_not_upgrade_completeness` **only** |

Mutations 3 and 4 red the same pair, so on mutation evidence alone the reconciliation control looks redundant against the compacted-summary one. **Mutation 6 is the separating case**, reddening it alone at `left: 4, right: 3`. What the extra control buys is proof that reconstruction genuinely *succeeds* in the fixture — so the degraded verdict is held **despite** working reconciliation rather than alongside broken reconciliation.

## Three fixture facts that cost a red test each

Recorded because the wrong version of each looks entirely reasonable:

- **`fold()` does not build `subject_summary`.** `fold_subject_summary()` is a separate call; a fixture running only the first produces **zero** summaries, and every subject-summary control would pass vacuously against a store that had never summarised anything.
- **The summary must be rebuilt AFTER compaction.** Folding first leaves the pre-compaction total in the row and the citation adds the removed event back on top of a count that never lost it — measured at `reconciled = 4` against a true 3. Green, and double-counting.
- **`is_admissible` drops `Expired` and `Inadmissible`, not staleness.** A stale claim is served by `ask` carrying a `Stale` grade, so the first draft of the history/`ask` contrast failed on its own premise; it now uses expiry.

## The 3e gate caught this PR's own fixture

`check_freshness_coverage.py` flagged the deliberately-unruled negative-control class — the gate working as designed. It's baselined as knowingly unruled, with the reasoning recorded: ruling it would delete the test's premise.

## Verification

- `cargo test -p kirra-world-store -p kirra-world-service` — 34 suites green, 0 failures
- `check_world_semantics`, `check_freshness_coverage`, `check_world_answer_boundary`, `check_kirra_world_bidirectional_fence`, `check_reexport_shims` — all green
- `cargo build --workspace`, `cargo fmt --all --check`, `cargo clippy --all-targets` on both crates — clean

With this, **3d (the typed query engine) is the sole remaining Tier 3 implementation box.**

---

### DIFF
--- a/ci/freshness_unruled_baseline.json
+++ b/ci/freshness_unruled_baseline.json
@@ -16,6 +16,11 @@
   ],
   "unruled": [
     {
+      "kind": "not_a_ruled_kind",
+      "predicate": "whatever",
+      "why": "The 3g follow-up's negative control. Deliberately names a class no ruling will ever cover, so `an_unruled_claim_refuses_the_whole_history` can prove that 3e's fail-closed refusal reaches the history family -- which is the ONLY reason `answer_admissibility` is in that family's version set. Ruling it would delete the test's premise."
+    },
+    {
       "kind": "observation",
       "predicate": null,
       "why": "A payload-only observation. `KIRRA-WM-CLAIM-SHAPES-001` makes this a legitimate shape, but what an untyped payload MEANS temporally cannot be ruled in general -- it depends on what the payload carries, which is precisely the thing no key can see."
--- a/crates/kirra-world-service/src/answer_ref.rs
+++ b/crates/kirra-world-service/src/answer_ref.rs
@@ -98,6 +98,22 @@ pub enum QueryKind {
     /// ([`crate::read_view::TemporalLookup::semantics`]), and that set has to
     /// be derived in the one place every set is derived.
     AsOfSubject,
+    /// [`crate::read_view::WorldView::history`] — every confirmed claim ever
+    /// recorded about one subject, in generation order.
+    ///
+    /// Distinct from [`Self::AsOfSubject`] despite both replaying the log: that
+    /// one answers *"what held at an instant"* and returns a SET, this one
+    /// answers *"what has ever been claimed"* and returns a SEQUENCE whose
+    /// order is part of the answer.
+    SubjectHistory,
+    /// [`crate::read_view::WorldView::subject_summary`] — the aggregate
+    /// summary for one subject, and how much of its evidence survives.
+    ///
+    /// The one family whose completeness is NOT a
+    /// [`kirra_world_store::Resolution`]. See
+    /// [`crate::read_view::SummaryLookup`] for why coercing it into one would
+    /// have to invent a field.
+    SubjectSummary,
     /// [`crate::lineage::LineageRef`] — one page of the evidence bearing on a
     /// subject at a pinned generation.
     ///
--- a/crates/kirra-world-service/src/read_view.rs
+++ b/crates/kirra-world-service/src/read_view.rs
@@ -98,6 +98,7 @@ use kirra_world::resolution::{resolve, RefusalReason, ResolutionOutcome};
 use kirra_world_store::compaction::Resolution;
 use kirra_world_store::entity_projection::IdentityView;
 use kirra_world_store::snapshot::SnapshotCoordinate;
+use kirra_world_store::subject_projection::SubjectSummaryAnswer;
 use kirra_world_store::{ProjectedClaim, StoreError, TrustAxes, TrustGrade, Validity, WorldStore};
 
 use crate::answer_ref::QueryKind;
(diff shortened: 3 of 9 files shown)


justinlooney:
Both addressed in `2c2697f8`. The first one is worth more than a fix note, because it is the same defect as last PR's.

## The per-subject scan (`read_view.rs:783`) — fixed, and it's a repeat

Correct, and sharper than it reads. `subject_summaries_with_coverage` is not merely *incidentally* whole-store — it is a **bulk API optimised as one**. Its own comment says the grouping exists *"because both tables accumulate for the life of a store, so the per-subject rescan this replaces was quadratic in exactly the dimension that never stops rising."* Standing a single-subject query on it inverts that optimisation precisely: I paid `O(total subjects + total citations)` to read one row.

**This is the same finding as #1440's unbounded lineage fetch** — same clause (`KIRRA-WM-ANSWER-IDENTITY-001` clause 2, *"queries are bounded"*), same invisibility (the returned answer was correct and bounded, so only the *work* was unbounded), one PR later, in a PR whose own description cites the #1440 write-up.

That recurrence is the useful part, and I've recorded it rather than just patching: **fixing the instance did not generalise.** A per-query bound is not a property anything currently checks, and both times it took a reviewer to see it. A gate that flags a bounded-looking boundary query built on an unbounded store call would be the real remedy; it is noted in `WM_SCOPE` as *not built*, rather than implied to be handled.

### The fix, in #1440's shape

`WorldStore::subject_summary_with_coverage` narrows both sources in SQL — `subject_summaries_for` and `load_summaries(Some(..))`, exactly as suggested. The part worth flagging beyond the suggestion: the coverage **verdict** is extracted into `subject_projection::coverage_from_citations`, so the bulk and narrowed paths run *the same code* rather than two implementations that agree today. A drifted coverage rule means one subject reported `Complete` by one path and `Degraded` by the other, which nothing else would catch.

`the_narrowed_and_bulk_coverage_paths_agree` sweeps every subject the bulk call knows — including the **citation-only** subject (`retained: None`), which a naive per-subject read drops and which is the case the bulk call was extended to stop losing — plus a subject in neither, for the `None` arm.

### Three mutations

| Mutation | Reds |
|---|---|
| narrowed path drops citation-only subjects | the agreement test |
| narrowed path reports `Complete` for a degraded subject | the agreement test |
| the **shared** rule stops matching `None`-kind citations | **four pre-existing tests** |

The third is what justifies the extraction: mutating the shared derivation reds the tests that guarded it while it was inline (`a_pre_migration_citation_still_degrades_its_subject` among them), so moving the rule out of the bulk function weakened nothing.

## The miscounted heading (`WM_SCOPE.md:3509`) — fixed

Correct. Both that heading and the test file's now say three. The test file also gains the bullet it was actually missing — *the summary must be rebuilt AFTER compaction* — which previously lived only on `holed_store`'s doc comment, so the header was both miscounted in one place and incomplete in the other.

## Verification

- `cargo test -p kirra-world-store -p kirra-world-service` — 34 suites green, 0 failures
- all five world CI gates green; `cargo build --workspace`, `cargo fmt --all --check`, `cargo clippy --all-targets` clean

---

PR #1477, opened 2026-08-22
KIRRA-WM-IDENTITY-AUTHORITY-001: every route to identity names an adjudicator

justinlooney:
Closes the finding 5c.1 surfaced.

> **An `IdentityAdjudication` may not be recorded without an authorized adjudicator.** `Merge`, `Split`, `Forget` and `Assert` each carry an `AdjudicationAuthority` on the **domain value**, so a constructed adjudication is always an authorized one.

## The gap, precisely

`AdjudicationAuthority` gated `same_as` promotion only. The other route to canonical identity — `append_adjudication`, which a `Merge` rides to `Lifecycle::Merged { into }` → the entity projection → identity resolution — required nothing.

The one constraint that *did* exist is SD-2: `append` refuses `LlmCandidate` writing `Confirmed`, so an LLM could not author an adjudication. **`Sensor` and `Derivation` could.** Two routes to the same effect, one admitting `Operator` only and the other admitting three classes.

## On the value, not the door

The alternative was a class check inside `append_adjudication`. Rejected because the domain value would still travel unauthorized: a second write door added later needs its own check, and nothing reminds the author. On the value, the refusal **cannot be reached around**.

What this does *not* buy is stated exactly, because the authority's own doc once overclaimed it and was corrected. The constructor accepts any `SourceClass` and refuses every one but `Operator` **at runtime**. It constrains the declared class, never the credential — so "operator only" is exactly as strong as the authentication on the write path, the AOU already recorded with `PROMOTION-001`. What the type buys is that the refusal cannot be forgotten downstream.

## The derived path inherits rather than mints

`promote_confirmed_same_as` carries the authority 2b already checked, so a derivation adds no adjudicator that was not authorized upstream. The instant and the authority are tracked in **one binding** — the adjudicator of the promotion that established the run is the same decision `began` names, and one binding stops *"when it became identity"* and *"who made it identity"* being answered by two different records after a later edit to one selection.

## The payload had to grow

An adjudicator that isn't persisted isn't an adjudicator — the decoder would have to invent one. Payload schema → **v2** with a required `adjudicator`; a v1 payload is refused through the existing `UnsupportedSchema` path rather than defaulted.

No migration accompanies the bump, and that's a claim about the data rather than an omission: `append_adjudication` has never had a production caller, so no deployed store can hold a v1 adjudication row.

## The pin did not fire

5c.1 left a fence documented as *"the thing that should fail and be rewritten"* when this ruling landed.

**It did not fail.** It asserted `RecordMerge` had grown no `authority` *field*; the ruling put the authority inside `MergeEntities` instead. The fence kept passing while the thing it guarded had already changed underneath it — a guard written against one shape of a change, satisfied by a different shape of the same change. Third time this tier a textual guard has been found shaped around the wrong axis.

The replacement is behavioural: **two merges identical but for who decided must produce different logs.** That fails if the command drops the authority, if the payload doesn't persist it, or if the encoding writes a constant — and no reshaping satisfies it by accident.

## Controls, each run as a mutation

| Mutation | Caught by |
|---|---|
| encoder writes a constant adjudicator | `two_merges_differing_only_in_adjudicator_are_different_records` |
| a v1 payload is read instead of refused | `a_v1_adjudication_payload_is_refused_rather_than_read_without_an_adjudicator` |
| a v2 payload with no adjudicator accepted | the same test's second half, which is also its non-vacuity control |

Constructing any verb without an authority is a **compile** error, which is why there's no test for it: a test that cannot be written in the language it tests is a test asserting the compiler works.

## Scope

`Assert` gained the requirement with the other three because the field is on the domain value and `IdentityAdjudication` has four arms. That does **not** unblock box 5d — operator teaching stays blocked, no command records an `Assert`, and `no_command_can_record_an_assert_because_5d_is_unruled` still pins it.

## Verification

- `cargo test --workspace` — **275 groups, 0 failed**
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets` — **zero** unused/dead-code warnings
- Twelve gates pass, including `check_world_semantics`

On the semantics gate: the corpus digest is **unmoved**. Authority does not reach the entity fold's output, so this is a source change with a held digest rather than a reducer change — verified by running the gate, not assumed.

Two process notes, since both cost time and one nearly shipped:

- A scripted call-site fixer was **not idempotent** and corrupted two files with repeated argument insertions. Caught by inspecting when the loop kept re-patching the same file, reverted, and redone deterministically. Every touched file was then audited for runaway insertions and duplicate helpers — all clean.
- One of those scripted edits left a fixture helper at file scope where it was dead code in a lib-only build. Moved inside its `cfg(test)` module.

Drive-by: removed the unused `StoreError` import I flagged in #1476 — I'm deep in that crate here, so it stopped being scope creep.

https://claude.ai/code/session_01XMCmUGL26v2Hk6eufnShxW

---


justinlooney:
`Branch Coverage (workspace)` is red, and it is **not this PR** — it fails identically on the base branch.

## What fails

```
error[E0282]: type annotations needed
  --> zenoh-1.9.0/src/lib.rs
error: could not compile `zenoh` (lib) due to 1 previous error
```

A third-party crate failing to compile, in its own source, under the **nightly** toolchain this lane uses for `llvm-cov`.

## Reproduced on the base branch

Checked out `28abc694` — plain `main`, none of this PR's changes — and ran the same compile with today's nightly:

```
$ git checkout 28abc694          # main
$ cargo +nightly check -p kirra-fleet-transport
error[E0282]: type annotations needed
error: could not compile `zenoh` (lib) due to 1 previous error
```

Identical failure, no changes of mine present.

## Why it started now

`rustc 1.100.0-nightly (c656540d6 2026-08-21)` published **2026-08-22**. Main's last CI run was 2026-08-21T23:44 and passed on the previous nightly, so this is the first run on the affected compiler.

This PR touches only `kirra-world*` and `kirra-proposal-context`. `zenoh` is a dependency of `kirra-fleet-transport`, reached from neither, and no change here can produce a type-inference error inside a vendored crate's own source.

## Status

Not fixing it in this PR — a nightly regression in a third-party crate is not this change's to carry, and the two obvious remedies (pinning the coverage lane's nightly, or bumping/patching `zenoh`) are repo-wide decisions that want their own PR and their own review.

Will re-check when the lane recovers. Every other check on this PR is green.

---

